PR #22, opened 2021-04-02
Completed Assignment 1.

delta-plus:
This is Joe Grassl, by the way.


delta-plus:
Forgot to do two PRs instead of one.

PR #46, opened 2021-04-04
Assignment 1 @anthonehi API

roylalog000:
pair coding with @anthonehi


SecretSanta/src/SecretSanta.Api/Startup.cs
```diff
@@ -32,7 +32,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             {
                 endpoints.MapGet("/", async context =>
                 {
-                    await context.Response.WriteAsync("Hello from API!");
+                    await context.Response.WriteAsync("Greetings from SecretSantaAPI!");
```
mmwoodfo, line 35: 🥇

PR #62, opened 2021-04-12
Assignment 2

Fervidusletum:


SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,49 @@
...
+                MockData.UpdateGift(viewModel);
+                return RedirectToAction(nameof(Index));
+            }
+            return View(viewModel);
+        }
+
+        [HttpPost]
+        public IActionResult Delete(int id)
+        {
+            MockData.Gifts.RemoveAt(id);
```
Thunderbird145, line 44: Website crashes when attempting to delete a gift

Thunderbird145: Users
Groups
Gifts
Title: ✔
Description: ✔
Url: ✔



All features are present, but gifts have a bug with deletion that needs fixed.

Fervidusletum: Missed updating that one, thanks for pointing it out!

SecretSanta/src/SecretSanta.Web/Data/MockData.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Data
+{
+    public class MockData
```
breaman, line 9: Very nice and clean way of making MockData more stand alone than what was done in class.

SecretSanta/src/SecretSanta.Web/Startup.cs
```diff
@@ -28,12 +29,13 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
             app.UseRouting();
 
+            // needed for what we're doing?
+            app.UseStaticFiles();
```
breaman, line 33: technically not needed till we care about pulling down static files (js and css) but not bad to add since we do add these files to our _Layout.cshtml, we just aren't using them yet.

SecretSanta/src/SecretSanta.Web/ViewModels/GiftViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        [Display(Name="Description")]
+        public string? Description { get; set; } = "";
+
+        [Display(Name="URL")]
+        [Url]
+        public string? URL { get; set; } = "";
+
+        [Display(Name = "Priority")]
+        [RegularExpression("^[0-9]+$")] // restrict to ints
```
breaman, line 29: Nice extra regex check

SecretSanta/src/SecretSanta.Web/Views/Gifts/Index.cshtml
```diff
@@ -0,0 +1,24 @@
+﻿@model List<GiftViewModel>
+
+@foreach (GiftViewModel gift in Model)
+{
+    <section style="margin-bottom: 1em">
+        <div>
+            Recipient: @MockData.GetUser(gift.UserId).FullName - <a asp-controller="Gifts" asp-action="Edit" asp-route-id=@gift.Id>Edit</a>
```
breaman, line 7: This works here, but when this gets pulled to using anything other than in memory data, you would probably want to have this all pulled in your view model instead so that you aren't making extra roundtrip calls in a foreach statement. I like the addition here, though.

SecretSanta/src/SecretSanta.Web/scripts/main.js
```diff
@@ -0,0 +1 @@
+import '../styles/site.css';
```
breaman, line 1: This file isn't needed since the .ts file is the only one that should be getting used.

breaman (commented):
Great job overall. I like the extra effort you put into the assignment to give the "ugly" pages some extra flash anyway.

PR #63, opened 2021-04-12
Assignment2 hw completed

Ashrom0:
Finished Assignment 2


TechGlue:
I will review this!

ybaghlaf (commented):
Users
Groups
Gifts
Title: ✔
Description: ✔
Url: ✔

SecretSanta/src/SecretSanta.Web/ViewModels/GiftViewModel.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace SecretSanta.Web.ViewModels
+{
+    public class GiftViewModel
+    {
+        public int Id { get; set; }
+        public string Title { get; set; } = "";
+        public string Description { get; set; } = "";
+        public string URL { get; set; } = "";
+        public int Priority { get; set; } = 1;  
+        public int UserId { get; set; }
```
TechGlue, line 13: For simplicity, we don't want to have preset values. It'd be best to allow the user to pick the priority of the Gift. I'm not too sure if this will show when creating a new gift but removing the default value of 1 would be best in case it does show and the user does not have to go through the process of deleting it.

SecretSanta/src/SecretSanta.Web/Data/MockData.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Collections.Generic;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Data
+{
+    public static class MockData
+    {
+        public static List<GiftViewModel> Gifts = new List<GiftViewModel>{
+            new GiftViewModel {Id = 0, Title = "Gift Title", Description = "Gift Description", URL ="Example.com", Priority = 1, UserId = 0},
```
TechGlue, line 10: This is up to you but maybe get rid of the preset values. If I was to use this application I would want the app to be blank so I don't have to go through the process of deleting preinstalled members.

TechGlue (commented):
Overall, it looks good. You have all the required materials for the site. But, I am having trouble building and running on my local machine so maybe spend some time to double-check that all the build files are in order. 

Users
Groups
Gifts
    Title: ✔
    Description: ✔
    Url: ✔

SecretSanta/src/SecretSanta.Web/Views/Gifts/Edit.cshtml
```diff
@@ -0,0 +1,7 @@
+@model GiftViewModel
+
+<form asp-controller="Gifts" asp-action="Edit" method="post">
+    @await Html.PartialAsync("_CreateOrEdit")
+    <button class="border border-blue-800 rounded-xl py-2 px-4 bg-blue-200 leading-10 mr-6"></button>Update</button>
```
breaman, line 5: You have an extra ending tag on the button which makes it so that the text "Update" is not clickable.

breaman (commented):
Looks like you are missing the _Layout_Template.cshtml file which makes it so this project doesn't even run unfortunately.

PR #64, opened 2021-04-12
Assignment 2 Completed 

TechGlue:


Ashrom0:
I will review this

TechGlue:
Not too sure the reasons why the build would not work from the usal commands but in order to get it running on your local machine. Need to do a "npm install" then a "npm run build:dev" and then you can do a "dotnet run" to get it built and running

Ashrom0 (commented):
Users
Groups
Gifts
Title: ✔
Description: ✔
Url: ✔

Nice work on the Assignment 2 hw

SecretSanta/src/SecretSanta.Web/Views/Home/Index.cshtml
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
breaman, line 1: This page should be using the _Layout file (which it is and is why you have navigation) but now in the body element of that page has html elements that it shouldn't.

breaman (commented):
Overall, nice work on the assignment, really only have one issue with it in the Home/Index.cshtml file.

PR #66, opened 2021-04-12
Completed Assignment 2.

delta-plus:
This is Joe Grassl. I'll try to get the next one in a bit sooner!

### DIFF
--- a/SecretSanta/src/SecretSanta.Web/Views/Gifts/Index.cshtml
+++ b/SecretSanta/src/SecretSanta.Web/Views/Gifts/Index.cshtml
@@ -1,0 +1,29 @@
+@{
+  ViewBag.Title = "SecretSanta";
+  int counter = 0;
+}
+
+<h1>Gifts</h1>
+
+<a asp-controller="Gifts" asp-action="Create">Create</a>
+
+@model List<GiftViewModel>
+
+@foreach(var gift in Model) {
+  <div>
+    @gift.Title - @gift.Description
+    <br>
+    Link: @gift.URL
+    <br>
+    Recipient ID: @gift.UserID
+    <br>
+    Priority: @gift.Priority
+    <div style="text-align: horizontal">
+     <a asp-controller="Gifts" asp-action="Edit" asp-route-id=@counter>Edit</a> 
+      <form asp-controller="Gifts" asp-action="Delete" asp-route-id=@counter method="post" style="display: inline-block">
+        <button>Delete</button>
+      </form>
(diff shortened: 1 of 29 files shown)


SecretSanta/src/SecretSanta.Web/Views/Gifts/Index.cshtml
```diff
@@ -0,0 +1,29 @@
...
+
+@model List<GiftViewModel>
+
+@foreach(var gift in Model) {
+  <div>
+    @gift.Title - @gift.Description
+    <br>
+    Link: @gift.URL
+    <br>
+    Recipient ID: @gift.UserID
```
Fervidusletum, line 18: You should present the name associated with the userid to the user, as the userid doesn't have much meaning to them.  
`Recipient ID: @TestData.Users[gift.UserID].FullName`

SecretSanta/src/SecretSanta.Web/Views/Groups/Create.cshtml
```diff
@@ -0,0 +1,9 @@
+@model GroupViewModel
+
```
Fervidusletum, line 2: You can condense most of the create and edit views into a single template.
[Example here](https://github.com/Fervidusletum/EWU-CSCD379-2021-Spring/tree/Assignment2/SecretSanta/src/SecretSanta.Web/Views/Groups)

Fervidusletum (approved):
- Create CRUD pages for the following components
- Users
- - A user only needs to have a First and Last name: ✔
- Groups
- - A group will only need to hold the name of the group: ✔
- Gifts
- - A gift should consist of the following information
- - Title: ✔
- - Description: ✔
- - Url: ✔
- - Priority (just a number for holding "order of importance"): ✔
- - The gift should hold a reference to the User it belongs to: ✔
- A Home page that can be used as just a standard landing page for the application: ✔
- Each page should have the ability to navigate between components: ✔

Looks good!

SecretSanta/src/SecretSanta.Web/Views/Groups/Create.cshtml
delta-plus: All good points. Thanks for the feedback!

breaman (commented):
Overall, nice work on the assignment. As your peer pointed out, could have saved yourself some duplicate effort by combining a lot of the functionality for the Create/Edit pages, but wasn't a necessity on this assignment, something to think about for future, though.

PR #68, opened 2021-04-12
Assignment2 started

jamesgreiner:


jamesgreiner:
Assignment 2 now finished per requirements in ReadME.md.

SecretSanta/src/SecretSanta.Web/Views/Groups/_CreateOrEdit.cshtml
```diff
@@ -0,0 +1,7 @@
+@model GroupViewModel
+
+<input type="hidden" asp-for="ID">
```
breaman, line 3: Having this here prevents someone from creating a group since the ID is required. This should have been moved to just be in the Edit.cshtml page.

SecretSanta/src/SecretSanta.Web/Views/Users/_CreateOrEdit.cshtml
```diff
@@ -0,0 +1,12 @@
+@model UserViewModel
+
+<input type="hidden" asp-for="ID">
```
breaman, line 3: Having this here prevents someone from creating a user since the ID is required. This should have been moved to just be in the Edit.cshtml page.

SecretSanta/src/SecretSanta.Web/Views/Users/_CreateOrEdit.cshtml
```diff
@@ -0,0 +1,12 @@
...
+<div>
+        <label asp-for="FirstName"></label>:<input type="text" asp-for="FirstName"/>
+        <span asp-validation-for="FirstName">First Name is required</span>
```
breaman, line 6: There should not be any text in here, it will get added automatically if things fail. By having the text in here already, it will always show.

breaman (commented):
Overall, not bad. Unfortunately I was unable to create a User or Group, but the Gifts worked just fine.

PR #69, opened 2021-04-12
Assignment 2

OkayNiceBrian:


SecretSanta/src/SecretSanta.Web/Data/MockData.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Data
+{
+    public static class MockData
+    {
+        public static List<GroupViewModel> Groups = new List<GroupViewModel>{
+            new GroupViewModel {Id = 0, GroupName = "First Group"},
+            new GroupViewModel {Id = 1, GroupName = "Second Group"},
```
CalumMackenzie-EWU, line 11: I like that you actually went out and instantiated values in your MockData for the GroupViewModel. It wasent needed but is good preparation for the future.

Users
Groups
Gifts
Title: ✔
Description: ✔
Url: ✔

SecretSanta/src/SecretSanta.Web/ViewModels/GiftViewModel.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace SecretSanta.Web.ViewModels
+{
+    public class GiftViewModel
+    {
+        public int Id {get; set;}
+        [Required]
+        public string Title {get; set;} = "";
+        public string Description {get; set;} = "";
+        public string Url {get; set;} = "";
+        public uint Priority {get; set;} = 1;
```
breaman, line 12: these technically could have just been integers, but I guess since they can't be negative numbers, a uint would work. Only concern would be in the future when we save to the database what would happen if the number got too big.

breaman (commented):
Nice work overall on the assignment. Good job getting a head start on the UI pieces as well.

PR #70, opened 2021-04-12
Assignment2

jmplett:


ZStuefen:
I will review this

ZStuefen:
# Assignment
- Create CRUD pages for the following components
  - Users
    - A user only needs to have a First and Last name: ✔
  - Groups
    - A group will only need to hold the name of the group: ✔
  - Gifts
    - A gift should consist of the following information
      - Title: ✔
      - Description: ✔
      - Url: ✔
      - Priority (just a number for holding "order of importance"): ✔
      - The gift should hold a reference to the User it belongs to: ❌
            - The user reference only holds a string, not a reference to the actual user
- A Home page that can be used as just a standard landing page for the application: ✔
- Each page should have the ability to navigate between components: ✔

breaman (commented):
Nice work on the assignment. One place to improve (but not necessary) was to combine the Edit and Create screens so you didn't need to duplicate so much of the html. Nice job listening to peer feedback as well and making corrections.

PR #71, opened 2021-04-12
Assignment 2 commit

Thunderbird145:
all requirements met

### DIFF
--- a/SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
+++ b/SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta name="viewport" content="width=device-width" />
+    <title>SecretSanta - @ViewBag.Title</title>
+    <!-- Stylesheets -->
+    <% for (var filename in htmlWebpackPlugin.files.css) { %>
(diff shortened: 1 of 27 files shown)


jmplett (commented):
Users
Groups
Gifts
Title: ✔
Description: ✔
Url: ✔

Great job everything seems to be in working order and meets the expectations.

SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
```diff
@@ -0,0 +1,32 @@
...
+    <title>SecretSanta - @ViewBag.Title</title>
+    <!-- Stylesheets -->
+    <% for (var filename in htmlWebpackPlugin.files.css) { %>
+    <link href="~<%= htmlWebpackPlugin.files.css[filename] %>" rel="stylesheet" />
+    <% } %>
+    <!-- End Stylesheets -->
+</head>
+
+<body>
+    @RenderBody()
```
breaman, line 14: Nothing wrong per se, but usually the navigation is at the top of the page instead of the bottom. Still meets the requirements, though.

breaman (commented):
cshtml filenames should be PascalCase but I didn't cover that specifically in class, so won't dock you for it. Nice work overall on the assignment, you hit all the elements that were asked for. One place for improvement, though, would have been to combine a lot of the markup for Create and Edit into a common file to prevent duplicate work.

PR #73, opened 2021-04-12
Assignment2

kaykay38:
Collaboration Tyler Jone @worseTyler and Mia Hunt @kaykay38


OkayNiceBrian:
I'll review this

SecretSanta/src/SecretSanta.Web/Data/MockData.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Data
+{
+    public static class MockData
+    {
+        public static List<UserViewModel> Users = new List<UserViewModel>
+        {
+            new UserViewModel {FirstName = "Tyler", LastName = "Jones", GroupName = "Super Duper Omega Astro Cool Giga Stomping Group", Id = 0, Gifts =
```
OkayNiceBrian, line 13: Nice way of doing this!

OkayNiceBrian (commented):
Users
Groups
Gifts
Title: ✔
Description: ✔
Url: ✔

Nicely done!

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,147 @@
...
+            return View(MockData.Users);
+        }
+        public IActionResult Create()
+        {
+            return View();
+        }
+
+        public IActionResult Update(int id)
+        {
+            MockData.Users[id].Id = id;
```
breaman, line 25: I found this was very "finicky" on order of how you did things to truly get all functionality to work. If you create a group on the group page, you can't edit your users anymore due to your associations you make in this method. Also I found if I created a group, then tried to add a user and associate them to that group, that also failed. Looks like it all boiled down to the fact that Group.Users was null when creating the group in the GroupController and that wreaked havoc with a lot of this logic in the UserController. Was still able to figure out how to do CRUD here with a user so not docking you since you did try to go above and beyond, but just things to be careful of as well.

breaman (commented):
Overall nicely done. You could have removed some duplicate html markup by combining a lot of that into a common view like what we did in class, but that wasn't necessary for this assignment so just mentioning it, not marking down for not having it. Liked the way you displayed the gifts page, that was slick.

PR #74, opened 2021-04-12
Micah Johnson assignment2

thecactus1:


cschadewitz:
Hey @thecactus1 ,
Heads up you have this pr pointed to the main branch instead of the assignment 2 branch. I made the same mistake on assignment 1.
Casey

SecretSanta/src/SecretSanta.Web/Controllers/HomeController.cs
```diff
@@ -0,0 +1,14 @@
+using Microsoft.AspNetCore.Mvc;
+
+namespace SecretSanta.Web.Controllers
+{
+    public class HomeController : Controller{
+        public ActionResult Index(){
+            return View();
+        }
+
+        public ActionResult Page(){
```
breaman, line 10: No view around this action, so probably should have removed it. There isn't a way to navigate to it, which is good, but still a bogus url that throws an unexpected error.

SecretSanta/src/SecretSanta.Web/Views/Shared/_ViewStart.cshtml
```diff
@@ -0,0 +1,3 @@
+@{
```
breaman, line 1: This file is a duplicate and doesn't need to exist.

SecretSanta/src/SecretSanta.Web/Views/Gifts/edit.cshtml
```diff
@@ -0,0 +1,26 @@
...
+        <label asp-for="URL"></label>: <input type="text" asp-for="URL" />
+        <span asp-validation-for="URL"></span>
+    </div>
+    <div>
+        <label asp-for="Priority"></label>: <input type="int" asp-for="Priority" />
+        <span asp-validation-for="Priority"></span>
+    </div>
+    <div>
+        <label asp-for="UserID"></label>: <input type="int" asp-for="UserID" />
+        <span asp-validation-for="UserID"></span>
```
breaman, line 23: Would have been nice to have this as a select field instead of freeform text since most of the time a person using your application won't know id's of other items.

breaman (commented):
Nice work overall. Could have saved yourself some html typing by combining the add/edit screens to a common file like we did in class, but no big deal. Also, it is best practice to name your views using PascalCasing but also realized I didn't go through that in class either, so just pointing that out.

PR #75, opened 2021-04-12
Assignment2

CalumMackenzie-EWU:
I have now got most things working. The pages navigate and things can be created, edited, and deleted. However my gifts section doesn't handle a new user being added to the list well. It can handle the reference to the user fine when its a hard coded one, but when a user is added it will mess things up. If your looking over my code please let me know if you see what the problem is.

*edit I found the problem but haven't found a fix yet. I found that when looking at the html via the inspect in the browser, that on the gifts edit page it would list it like:

select data-val="true" data-val-required="The Suggested For field is required." id="UserId" name="UserId"
            Please select one 
option selected="selected" value="0" Rob Boss option
option value="1"Rick Vames option
option selected="selected" value="0" Mick Thames option
select

Which shows that two options are being selected as the users for the gift.
Note its not properly in html since html wont show up in the comment.
*end edit.


MeguminXIII:
I will Review

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,69 @@
...
+            return View(MockData.Gifts[id]);
+        }
+
+        [HttpPost]
+        public IActionResult Edit(GiftViewModel vModel)
+        {
+            if(ModelState.IsValid)
+            {
+                //Gifts[vModel.Id] = vModel;
+                MockData.Gifts[vModel.Id] = vModel;
```
MeguminXIII, line 56: So what I did was instead of this, I did MockData.GiftsList[vm.GiftId][vm.ID] = vm;, basically instead of using only one ID I used both the user Id and the Gift id.

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,69 @@
...
+            //MockData.Gifts[id].Id = id;
+            return View(MockData.Gifts[id]);
+        }
+
+        [HttpPost]
+        public IActionResult Edit(GiftViewModel vModel)
+        {
+            if(ModelState.IsValid)
+            {
+                //Gifts[vModel.Id] = vModel;
```
MeguminXIII, line 55: This comment doesn't really add anything so you can probably delete it.

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,69 @@
...
+            return View(vModel);
+        }
+
+        public IActionResult Edit(int id)
+        {
+            //Gifts[id].Id = id;
+            //return View(Gifts[id]);
+
+
+            //MockData.Gifts[id].Id = id;
```
MeguminXIII, line 46: These comments don't really add anything so you can probably remove them.

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,69 @@
...
+        public IActionResult Create()
+        {
+            return View();
+        }
+        [HttpPost]//cal:This maps the below version to the html method="post"
+        public IActionResult Create(GiftViewModel vModel)
+        {
+            if(ModelState.IsValid)//cal:if data is invalid goes back to create page, but also preserves data.
+            {
+                //Gifts.Add(vModel);
```
MeguminXIII, line 33: This comment can probably be removed as well

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public class GiftsController : Controller
+    {
+        //cal:This stuff gets moved to the Data/MockData.cs file so it can be referenced by other classes. Note some references must change to compensate.
+        /*
+        static List<GiftViewModel> Gifts = new List<GiftViewModel>{
+            new GiftViewModel{Title = "Lootbox", Description = "Let your friends know that you would rather leave their gift up to chance rather than give a bad gift.", Url="https://www.ea.com", Priority = 4},
+            new GiftViewModel{Title = "Giftcard", Description = "For when you either dont know your friend enough. Or know them a little too well.", Url="http://NotASketchyShop.com.", Priority = 1}
+        };
+        */
```
MeguminXIII, line 16: Can probably get rid of this comment as well. doesn't add anything.

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,69 @@
...
+                //Gifts[vModel.Id] = vModel;
+                MockData.Gifts[vModel.Id] = vModel;
+                return RedirectToAction(nameof(Index));
+            }
+            return View(vModel);
+        }
+
+        public IActionResult Delete(int id)
+        {
+            //Gifts.RemoveAt(id);
```
MeguminXIII, line 64: Can probably remove this as well

SecretSanta/src/SecretSanta.Web/Controllers/HomeController.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            return View();
+        }
+
+        /*
+        public IActionResult Page()
+        {
+            return View();
+        }
+        */
```
MeguminXIII, line 17: can probably remove this comment

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    public class UsersController : Controller
+    {
+        //cal:This stuff gets moved to the Data/MockData.cs file so it can be referenced by other classes. Note some references must change to compensate.
+        /*
+        static List<UserViewModel> Users = new List<UserViewModel>{
+            new UserViewModel{FirstName = "Rob", LastName = "Boss"},
+            new UserViewModel{FirstName = "Rick", LastName= "Vames"}
+        };
+        */
```
MeguminXIII, line 16: can probably remove this comment

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,66 @@
...
+        public IActionResult Index()
+        {
+            //return View();
+            //return View(Users);
```
MeguminXIII, line 20: Can probably remove this comment as well

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,66 @@
...
+        public IActionResult Create()
+        {
+            return View();
+        }
+        [HttpPost]//cal:This maps the below version to the html method="post"
+        public IActionResult Create(UserViewModel vModel)
+        {
+            if(ModelState.IsValid)//cal:if data is invalid goes back to create page, but also preserves data.
+            {
+                //Users.Add(vModel);
```
MeguminXIII, line 33: This comment doesn't really add anything so you can probably remove it

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+            return View(vModel);
+        }
+
+        public IActionResult Edit(int id)
+        {
+            //Users[id].Id = id;
+            //return View(Users[id]);
+            
+            //MockData.Users[id].Id = id;
```
MeguminXIII, line 45: This comment doesn't really add anything so you can probably remove it

SecretSanta/src/SecretSanta.Web/Startup.cs
```diff
@@ -28,12 +31,21 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+
             app.UseEndpoints(endpoints =>
             {
+                /*cal:removing this default. Is used to set things up manually
                 endpoints.MapGet("/", async context =>
                 {
                     await context.Response.WriteAsync("Hello from Web!");
                 });
+                //end default
+                */
```
MeguminXIII, line 44: Can probably remove this as well

MeguminXIII (commented):
Assignment
Users
Groups
Gifts
Title: ✔
Description: ✔
Url: ✔

looks pretty good. for the gifts section adding new users, I don't see anything about users in the /Views/Gifts/Index.cshtml, though I think that would just affect the presentation of the list as opposed to the function of the list. I found one place but I will put that in a separate comment.

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
CalumMackenzie-EWU: Thank you for the thorough review with all the stuff. Yeah I should clean up many of the old comments. I leave them there so when I go back to them at a later date I can see what the default values were before they were changed for the specific assignment. As for the possible solution, I'm trying out your suggestion and it sadly hasn't fixed things, but thank you for the it nonetheless. I figured out what's wrong just currently don't know how to solve it.

SecretSanta/src/SecretSanta.Web/Views/Users/Index.cshtml
```diff
@@ -0,0 +1,16 @@
+<h1>Welcome to the Users Page</h1>>
```
breaman, line 1: Looks like you have an extra '>' at the end of the line here. No big deal, just pointing it out so you can clean it up.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Create.cshtml
```diff
@@ -0,0 +1,22 @@
...
+    </div>
+    <div>
+        <label asp-for="Url"></label>: <input type="text" asp-for="Url" />
+        <span asp-validation-for="Url"></span>
+    </div>
+    <div>
+        <label asp-for="Priority"></label>: <input type="text" asp-for="Priority" />
+        <span asp-validation-for="Priority"></span>
+    </div>
+    <button type="submit">Create</button>
```
breaman, line 20: Why on creation can I not associate a user to a gift? Noticed I can do it on edit, but not here. This is one place where doing what we went through in class and combining your Create/Edit screens into a common cshtml file comes in handy so that you don't make copy/paste errors and remember to have to change both when you make a change to the other.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Edit.cshtml
```diff
@@ -0,0 +1,28 @@
...
+        <label asp-for="Priority"></label>: <input type="text" asp-for="Priority" />
+        <span asp-validation-for="Priority"></span>
+    </div>
+    <div>
+        <label asp-for="UserId"></label>
+        <select asp-for="UserId" asp-items="@(new SelectList(MockData.Users,"Id","FullName"))">
+            <option>Please select one</option>
+        </select>
+    </div>
+    <button type="submit">Edit</button>
```
breaman, line 26: Not docking for this, but renaming this something like "Save" is a little more intuitive for a user. When I see something that says "Edit", I don't expect it to persist my results, but would expect it to take me somewhere so that I can edit things.

breaman (commented):
Nice work overall, the application worked well and hit the requirements.

PR #76, opened 2021-04-12
Assignment2

Cybotive:


cschadewitz:
I'll review this.

SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Controllers
+{
+    public class GiftsController : Controller
+    {
+        static List<GiftViewModel> _Gifts = new List<GiftViewModel>()
```
cschadewitz, line 10: I am sure this was done out of convenience but you could segment the responsibility of data retention away from the controller.

SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public IActionResult Create()
+        {
+            return View();
+        }
+
+        [HttpPost]
+        public IActionResult Create(GroupViewModel viewModel)
+        {
+            if (ModelState.IsValid)
```
cschadewitz, line 29: It is possible for ModelState.IsValid to be true even though the model is null and although this isnt intended to be an API you could add a null check.

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Controllers
+{
+    public class UsersController : Controller
+    {
+        static List<UserViewModel> _Users = new List<UserViewModel>()
```
cschadewitz, line 10: Good job with naming specifications

SecretSanta/src/SecretSanta.Web/ViewModels/GiftViewModel.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        [Required]
+        [Display(Name = "Title")]
+        public string Title { get; set; } = "";
+        [Required]
+        [Display(Name = "Description")]
+        public string Desc { get; set; } = "";
+        [Required]
+        [Display(Name = "URL")]
+        public Uri? Url { get; set; }
```
cschadewitz, line 19: Does it make sense for this field to be required and nullable?

SecretSanta/src/SecretSanta.Web/ViewModels/GiftViewModel.cs
```diff
@@ -0,0 +1,25 @@
...
+        [Display(Name = "Priority")]
+        public int Priority { get; set; } = -1;
+        [Display(Name = "Recipient")]
+        public UserViewModel? Recipient { get; set; }
```
cschadewitz, line 23: It would make sense to me for every gift to have an associated recipient, so I would think this should be required. Additionally, I would think you would want to retain an ID of the recipient rather than a reference to the user.
Does this work properly with the <select> element?

SecretSanta/src/SecretSanta.Web/styles/site.css
```diff
@@ -0,0 +1,3 @@
+﻿@tailwind base;
```
cschadewitz, line 1: I believe that you need to remove the content from site.css because you havent included a tailwind.config.js and postcss.config.js

cschadewitz (commented):
Overall: Looks good


        Users
        Groups
        Gifts
                Title: ✔
                Description: ✔
                Url: ✔

Note:
I wasn't able to get the webpack build working.
Initially I believed it was due to missing tailwind.config.js and postcss.config.js. 
However, I believe it is an issue on my side as I havent been able to get the setup from Stokes to build on my machine either.
I will further confirm functionality upon getting the webpack build to work.

SecretSanta/src/SecretSanta.Web/ViewModels/GiftViewModel.cs
Cybotive: Good catch. I was probably dealing with a warning without thinking.

SecretSanta/src/SecretSanta.Web/ViewModels/GiftViewModel.cs
Cybotive: I interpreted the specification quite literally. Using ID would definitely be more simple.

SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
Cybotive: Thanks for the info. My eye was drawn to this potential issue but I wasn't sure how ModelState behaved.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Create.cshtml
```diff
@@ -0,0 +1,28 @@
...
+        <label asp-for="Url"></label>: <input type="text" asp-for="Url" />
+        <span asp-validation-for="Url"></span>
+    </div>
+    <div>
+        <label asp-for="Priority"></label>: <input type="text" asp-for="Priority" />
+        <span asp-validation-for="Priority"></span>
+    </div>
+    <div>
+        <label asp-for="Recipient"></label>: 
+        <select asp-for="Recipient">
```
breaman, line 22: you aren't populating this like we did in class. You are missing the asp-items tag that would allow your users to populate this list. Without this, I'm unable to associate a user to a gift.

breaman (commented):
Overall nice work on the app. Unfortunately you were missing the piece that prevented me from being able to associate a user to a gift, but did great everywhere else.

PR #77, opened 2021-04-12
Assignment2

MeguminXIII:
Sorry it's late thought it was due today not yesterday.
It's also unfinished


delta-plus:
I will review this.

delta-plus (commented):
Nice work! Looks good.

        Users
            A user only needs to have a First and Last name:  ✔
        Groups
            A group will only need to hold the name of the group:  ✔
        Gifts
                Title: ✔
                Description: ✔
                Url: ✔
    A Home page that can be used as just a standard landing page for the application:  ✔
    Each page should have the ability to navigate between components:  ✔

delta-plus:
Solid commits! Good job fixing things up.

SecretSanta/dist/main.js
```diff
@@ -0,0 +1,15 @@
+/*
```
breaman, line 1: not sure what this file is, but I know we don't need it.

SecretSanta/src/SecretSanta.Web/Data/MockData.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Data
+{
+    public static class MockData
+    {
+        public static List<GroupViewModel> Groups = new List<GroupViewModel>
+        {
+            new GroupViewModel {GroupName = Users![0].GroupName!, Users = new List<UserViewModel>{Users[0], Users[1]}}
```
breaman, line 13: Since Users hasn't been defined yet, your application doesn't even run with this like this. I manually moved the Groups initializer to the end and got things to work, but I should not have had to do that.

SecretSanta/src/SecretSanta.Web/Controllers/GiftControllers.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace SecretSanta.Web.Controllers
+{
+    public class GiftsController : Controller
+    {
+        public IActionResult Create(){
+            return View();
+        }
+
+        public IActionResult Index(){
+            return View(MockData.GiftsList);
```
breaman, line 18: You are passing in Gifts, but your view is expecting Users. This crashes the app and prevents me from being able to do anything within Gifts without making modifications.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Edit.cshtml
```diff
@@ -0,0 +1,27 @@
+@model GiftViewModel
+
+<form asp-controller="Gifts" asp-action="Update" method="post">
```
breaman, line 3: Your action here for the Edit is pointing back to an Update method which doesn't exist. This should have pointed back to Edit instead.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Create.cshtml
```diff
@@ -0,0 +1,27 @@
+@model GiftViewModel
+
+<form asp-controller="Gifts" asp-action="Update" method="post">
```
breaman, line 3: This action is pointing back to an Update method which doesn't exist and should probably be pointing back to your Create method instead.

breaman (commented):
Looks like you have a couple of erroneous main.js files lying around here in this PR. Didn't dock you for it, but they aren't needed so should have been cleaned up.

PR #82, opened 2021-04-14
Assignment 2

NicholasJohnsonDev:
Sorry that this is late, won't let it happen again.

- Nicholas Johnson


thecactus1 (approved):
        Users
        Groups
        Gifts
                Title: ✔
                Description: ✔
                Url: ✔

SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE HTML>
```
breaman, line 1: This doesn't quite look right since it should have had the hooks in place for adding webpack assets. I also noticed, though, that you didn't have webpack configured so I had to manually copy this file and rename it to _Layout.cshtml in order to even get anything to work.

breaman (commented):
Once I got the _Layout file added, everything looked really nice. Only suggestion I would have had was to combine your Create/Edit cshtml files like we did in class just to make it so you didn't have to duplicate so much code. Not a big deal in a small project like this, but can be an issue when you get larger forms.

PR #85, opened 2021-04-19
Assignment 3 Micah Johnson

thecactus1:


OkayNiceBrian:
I will review this

SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
```diff
@@ -18,6 +18,22 @@
             <li><a asp-controller="Gifts" asp-action="Index">Gifts</a></li>
         </ul>
     </nav>
+<body>
```
OkayNiceBrian, line 21: This body tag is an error, your assignment didn't compile until I deleted it.

OkayNiceBrian (commented):
The navigation elements should be layed out horizontally instead of vertically: ✔(although they are still layed out vertically at the top)
Instead of "Home" the text for that link should be Secret Santa and the font for this should be a bit larger than the other navigation elements: ✔ (Although Home still exists)
The page content should have margin around the whole thing so that the text isn't butted up right against the edge of the screen: ❌
The user "home" page should
When on a medium or larger screen, display users in rows of no more than 3 users per row: ❌
Clicking the user on either screen should take the person to an edit screen: ❌ (edit link still exists)
There should be an "icon" of some sort on each "tile" that when clicked will delete the user: ❌ (still an X link, not a fontawesome (for example) icon)
When deleting, the person should be prompted with a confirmation dialog before actually performing the delete: ❌
The gift "home" page should
The gifts should be ordered by priority, where priority 1 items are at the top: ❌
When on a medium or larger screen, display gifts in rows of no more than 3 gifts per row (Title only): ❌
Clicking the gift on either screen should take the person to an edit screen: ❌
There should be an "icon" of some sort on each "tile" that when clicked will delete the gift: ❌
When deleting, the person should be prompted with a confirmation dialog before actually performing the delete: ❌
The user "edit" page should
When on a medium or larger screen, the first name and last name editing should be on the same line (side by side): ❌
The "save" and "cancel" buttons should have some type of hover effect so the person knows when their cursor is over the button: ❌
The gift "edit" page should
When on a medium or larger screen, the title and description should be vertically aligned, but the url and priority should be layed out side by side: ❌
If there is a data issue with any of the text boxes upon submit, then when the screen is presented again, the input box should reflect an error has occurred: ❌
The "save" and "cancel" buttons should have some type of hover effect so the person knows when their cursor is over the button: ❌
Make it so that the Url and Description fields are not required: ❌
Fix the title on every page to reflect this is a Secret Santa app instead of a UserGroup app: ❌

SecretSanta/src/SecretSanta.Web/Views/Gifts/_CreateOrEdit.cshtml
```diff
@@ -8,12 +8,10 @@
 <div>
```
breaman, line 8: you forgot to make this responsive based on screen size.

SecretSanta/src/SecretSanta.Web/Views/Users/Create.cshtml
```diff
@@ -1,7 +1,10 @@
 @model UserViewModel
 
-<form asp-controller="Users" asp-action="Create" method="post">
-    @await Html.PartialAsync("_CreateOrEdit")
-    <button>Create</button>
-    <a asp-controller="Users" asp-action="Index">Cancel</a>
-</form>
\ No newline at end of file
+ViewBag.Title = "User Create";}
```
breaman, line 3: missed the @{ piece of code on this one so this actually showed up in your html.

SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
```diff
@@ -2,22 +2,29 @@
...
-    </nav>
+    <header class="bg-blue-500 text-white flex flex-col sm:flex-row justify-between p-6" x-data="SecretSanta.setupNav()">
+        <div class="flex flex-row justify-between">
+            <a class="text-2xl leading-10" asp-controller="Home" asp-action="Index">Secret Santa</a>
+            @*<button class="sm:hidden" onclick="SecretSanta.toggleMenuWithJs()"><i class="fas fa-bars text-2xl align-middle h-full"></i></button>*@
+            <button class="sm:hidden" onClick="SecretSanta.toggleMenuWithJs()"><i class="fas fa-bars text-2xl align-middle h-full"></i></button>
+        </div>
+        <nav id="headerNav" class="hidden sm:inline-block">
+            <ul class="flex flex-col sm:flex-row">
+                <li class="sm:px-10 leading-10"><a asp-controller="Home" asp-action="Index"><b>HOME</b></a></li>
```
breaman, line 21: This link could have been deleted since the Secret Santa link takes you to the same place.

breaman (commented):
Nice work overall on the assignment, this looked pretty good for the most part.

PR #86, opened 2021-04-19
Assignment 3 completed

ZStuefen:


jamesgreiner:
I will review this.

jamesgreiner (commented):
- The header should have the following elements
  - The header should have a background color and should extend the full width of the screen: ✔
  - The navigation elements should be layed out horizontally instead of vertically: ✔
  - The items should not look like a bulleted list, but should look more like a typical navigation bar: ✔
  - Instead of "Home" the text for that link should be Secret Santa and the font for this should be a bit larger than the other navigation elements: ✔
  - The navigation should be responsive, which means on a small screen the navigation should be hidden until the user clicks on some type of indicator to expand out the navigation elements: ✔
- The page content should have margin around the whole thing so that the text isn't butted up right against the edge of the screen: ✔
- The user "home" page should
  - When on a small screen, display all users in a single column (First and Last name): ✔
  - When on a medium or larger screen, display users in rows of no more than 3 users per row: ✔
  - Clicking the user on either screen should take the person to an edit screen: ✔
  - There should be an "icon" of some sort on each "tile" that when clicked will delete the user: ✔
    - When deleting, the person should be prompted with a confirmation dialog before actually performing the delete: ✔
- The gift "home" page should
  - The gifts should be ordered by priority, where priority 1 items are at the top: ✔
  - When on a small screen, display all gifts in a single column (Title only): ✔
  - When on a medium or larger screen, display gifts in rows of no more than 3 gifts per row (Title only): ✔
  - Clicking the gift on either screen should take the person to an edit screen: ✔
  - There should be an "icon" of some sort on each "tile" that when clicked will delete the gift: ✔
    - When deleting, the person should be prompted with a confirmation dialog before actually performing the delete: ✔
- The user "edit" page should
  - When on a small screen, fields should be lined up in a single column: ✔
  - When on a medium or larger screen, the first name and last name editing should be on the same line (side by side): ✔
  - If there is a data issue with any of the text boxes upon submit, then when the screen is presented again, the input box should reflect an error has occurred: ✔
  - The "save" and "cancel" buttons should have some type of hover effect so the person knows when their cursor is over the button: ✔
- The gift "edit" page should
  - When on a small screen, fields should be lined up in a single column: ✔
  - When on a medium or larger screen, the title and description should be vertically aligned, but the url and priority should be layed out side by side: ✔
  - If there is a data issue with any of the text boxes upon submit, then when the screen is presented again, the input box should reflect an error has occurred: ✔
  - The "save" and "cancel" buttons should have some type of hover effect so the person knows when their cursor is over the button: ✔
  - Make it so that the Url and Description fields are not required: ✔
- Fix the title on every page to reflect this is a Secret Santa app instead of a UserGroup app: ✔


Everything appears to meet the assignment requirements. It loads up fine in a browser and looks clean and simple. Nice work!

SecretSanta/src/SecretSanta.Web/Views/Gifts/Create.cshtml
```diff
@@ -1,7 +1,9 @@
-@model GiftViewModel
+﻿@model GiftViewModel
 
-<form asp-controller="Gifts" asp-action="Create" method="post">
-    @await Html.PartialAsync("_CreateOrEdit")
-    <button>Create</button>
-    <a asp-controller="Gifts" asp-action="Index">Cancel</a>
-</form>
+<section class="m-5">
```
breaman, line 3: You should have put this section tag in the _Layout_Template.cshtml file so that you would not have had to add it to every file instead.

SecretSanta/src/SecretSanta.Web/Views/Gifts/_CreateOrEdit.cshtml
```diff
@@ -1,28 +1,47 @@
...
+<div class="flex mb-6">
+    <div class="flex w-1/2">
+        <div class="w-1/3">
+            <label class="text-gray-500 font-bold text-right" asp-for="Url"></label>
+        </div>
+        <div class="w-2/3">
+            <input class="focus:outline-none focus:ring-transparent focus:border-indigo-600" type="text" asp-for="Url" />
+        </div>
+    </div>
+    <div class="flex w-1/2">
```
breaman, line 32: This should have been lined up under the Url on a small screen.

breaman (commented):
Nice work overall on the assignment. Missed one small element, but nothing major.

PR #89, opened 2021-04-19
Assignment 3 completed

Ashrom0:
Note: I renamed the file "_Layout_Template" to "_Layout" to I can run it as i was having trouble running nmp install and npm build:dev.


TechGlue:
I'll review this!

TechGlue (commented):
The navigation elements should be layed out horizontally instead of vertically: ❌
The items should not look like a bulleted list, but should look more like a typical navigation bar: ❌
The user "home" page should
The gift "home" page should
The gifts should be ordered by priority, where priority 1 items are at the top: ❌
The user "edit" page should
When on a medium or larger screen, the first name and last name editing should be on the same line (side by side): ✔
If there is a data issue with any of the text boxes upon submit, then when the screen is presented again, the input box should reflect an error has occurred: ❌
The gift "edit" page should
When on a medium or larger screen, the title and description should be vertically aligned, but the url and priority should be layed out side by side: ✔
Make it so that the Url and Description fields are not required: ✔

TechGlue:
It's looking good but there are some edits you can make. Users, the edit is not functioning correctly with blank names being able to pass through. Space out the elements a bit to make it look cleaner but not required. 

Gifts are missing a create button or I did not see it when I did my review. The fields in the gifts edit could also use some editing to make it look a bit cleaner but not required. When I add blank edits in gifts it lets me pass through and does not warn me that the field is required.

Overall, you got the main view and functionality of the site going for you just a few additions are needed!

SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
```diff
@@ -1,29 +1,51 @@
...
-    <meta name="viewport" content="width=device-width" />
-    <title>UserGroup - @ViewBag.Title</title>
+    <meta name="viewport" content="width=device-width"/>
+    <title>SecretSanta - @ViewBag.Title</title>
     <!-- Stylesheets -->
-    <% for (var filename in htmlWebpackPlugin.files.css) { %>
-    <link href="~<%= htmlWebpackPlugin.files.css[filename] %>" rel="stylesheet" />
-    <% } %>
+
+    <link href="~/css/main.6c5ef2f7ce93b0128af7.css" rel="stylesheet" />
```
breaman, line 8: Why was this hardcoded instead of using the code we had originally that uses the generated files instead? Doing this completely broke your UI.

SecretSanta/src/SecretSanta.Web/styles/site.css
```diff
@@ -0,0 +1,20 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+
```
breaman, line 4: Looks like you forgot to add the font awesome elements in here which made it so none of your icons were displayed.

breaman (commented):
I had to make a change to your _Layout_Template file to even get this assignment to look right. Also looks like you missed out on configuring font-awesome so I couldn't even test a few pieces.

PR #90, opened 2021-04-19
Assignment3

OkayNiceBrian:


CalumMackenzie-EWU:
Starting review.

SecretSanta/src/SecretSanta.Web/Views/Users/_CreateOrEdit.cshtml
```diff
@@ -1,12 +1,9 @@
...
-    <input type="text" asp-for="FirstName" />
-    <span asp-validation-for="FirstName"></span>
-</div>
-<div>
-    <label asp-for="LastName"></label>
-    <input type="text" asp-for="LastName" />
-    <span asp-validation-for="LastName"></span>
-</div>
\ No newline at end of file
+<section class="m-2">
```
CalumMackenzie-EWU, line 2: The user "edit" page should
When on a medium or larger screen, the first name and last name editing should be on the same line (side by side): ❌
I may be misunderstanding the brief. But I think we are supposed to have both the first and last name appear side by side when its medium or larger. Should be a quick fix to just move them to their own columns.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Edit.cshtml
```diff
@@ -1,8 +1,11 @@
 @model GiftViewModel
+@{
+    ViewBag.Title = "Edit Gift";
+}
 
 <form asp-controller="Gifts" asp-action="Edit" method="post">
     <input type="hidden" asp-for="Id" />
     @await Html.PartialAsync("_CreateOrEdit")
```
CalumMackenzie-EWU, line 8: This isn't the exact spot but just putting it here to point towards the fix should go inside the _CreateOrEdit file for Gifts.
The gift "edit" page should
When on a medium or larger screen, the title and description should be vertically aligned, but the url and priority should be layed out side by side: ❌
Just in case you're having trouble with this you can use the md: prefix for medium or larger screens the same way we used sm: in class.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Create.cshtml
```diff
@@ -1,7 +1,10 @@
 @model GiftViewModel
```
CalumMackenzie-EWU, line 1: This isnt the exact spot. This is just to point that the problem I believe is inside the GiftViewModel. It might also be in the controller but Im pretty sure its in the view model.
The gift "edit" page should
Make it so that the Url and Description fields are not required: ❌

CalumMackenzie-EWU (commented):
The navigation elements should be layed out horizontally instead of vertically: ✔
The user "home" page should
The gift "home" page should
The gifts should be ordered by priority, where priority 1 items are at the top: ✔
The user "edit" page should
When on a medium or larger screen, the first name and last name editing should be on the same line (side by side): ❌
The gift "edit" page should
When on a medium or larger screen, the title and description should be vertically aligned, but the url and priority should be layed out side by side: ❌
Make it so that the Url and Description fields are not required: ❌
Fix the title on every page to reflect this is a Secret Santa app instead of a UserGroup app: ✔

This is the overview of the review. Ive left specifics in some of the lines if possible.
I really like a lot of the things you got working like getting the nav element to react to actually resizing and not just the bars icon. And properly using ViewBag to title your stuff.

breaman (commented):
Unfortunately you missed the items that were pointed out in the PR and never corrected. Also noticed that you did not order the gifts based on priority.

PR #92, opened 2021-04-19
Assignment3 with @anthonehi

roylalog000:
pair coding with @anthonehi


OkayNiceBrian:
I will review this

SecretSanta/src/SecretSanta.Web/Views/Groups/Index.cshtml
```diff
@@ -4,8 +4,22 @@
...
     <section>
-        @groupViewModel.Name <a asp-action="Edit" asp-route-id=@counter>Edit</a><form asp-action="Delete" asp-route-id=@counter method="post"><button>X</button></form>
+        @groupViewModel.Name <a asp-action="Edit" asp-route-id=@counter>Edit</a><form asp-action="Delete" asp-route-id=@counter method="post"><button onclick="GetConfirmation()">X</button></form>
         @{counter++;}
     </section>
 }
+<script type="text/javascript">
+  function GetConfirmation()
+  {
+      var reply = confirm("Are you sure you want to delete this?");
```
breaman, line 14: Like stated in the earlier function around getting confirmation, this could have been written in just one line.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Edit.cshtml
```diff
@@ -3,6 +3,86 @@
...
+
+.button:hover span {
+  padding-right: 25px;
+}
+
+.button:hover span:after {
+  opacity: 1;
+  right: 0;
+}
+</style>
```
breaman, line 49: style tags should be at the top of the page instead of the bottom since if you have a lot of data to load, you don't want the page to load and then get styled after the fact. Why two different style section? Also, this was a lot of copy/paste, why not style a "button" and then just change the certain elements that were different? The only difference between these two buttons was the actual background-color item.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Index.cshtml
```diff
@@ -1,11 +1,80 @@
...
 
 @{int counter = 0;}
-
+<div class = "grid">
+<div>
 @foreach(var groupViewModel in Model) {
-    <section>
-        @groupViewModel.Title <a asp-action="Edit" asp-route-id=@counter>Edit</a><form asp-action="Delete" asp-route-id=@counter method="post"><button>X</button></form>
+    <section class = "grid">
+         <a asp-action="Edit" asp-route-id=@counter>@groupViewModel.Title</a><form asp-action="Delete" onclick ="return(confirmdelete())" asp-route-id=@counter method="post"><button class = "button"><span>X</span></button></form>
```
breaman, line 8: The assignment specifically called out that an icon should have been used instead of just an "x". This is where font-awesome should have been used like what was demonstrated in class.

breaman (commented):
Unfortunately without some tweaking, I wasn't even able to get this assignment to run since the _Layout_Template.cshtml file was deleted. I know it wasn't specifically stated in the assignment so I won't dock you for using pure css instead of tailwind, the assumption was you would have used the tailwind library that was covered in lecture.

PR #94, opened 2021-04-19
Assignment3

jmplett:


Cybotive:
I will review this.

SecretSanta/src/SecretSanta.Web/Views/Users/Create.cshtml
```diff
@@ -1,7 +1,9 @@
+@{ViewBag.Title = "Create User";}
+
 @model UserViewModel
 
 <form asp-controller="Users" asp-action="Create" method="post">
     @await Html.PartialAsync("_CreateOrEdit")
-    <button>Create</button>
-    <a asp-controller="Users" asp-action="Index">Cancel</a>
+    <button class="btn">Create</button>
```
Cybotive, line 7: Fast and simple.

SecretSanta/src/SecretSanta.Web/Views/Users/Create.cshtml
```diff
@@ -1,7 +1,9 @@
+@{ViewBag.Title = "Create User";}
```
Cybotive, line 1: Descriptive Titles across the app.

SecretSanta/src/SecretSanta.Web/Views/Gifts/_CreateOrEdit.cshtml
```diff
@@ -1,28 +1,30 @@
...
-    <span asp-validation-for="Priority"></span>
-</div>
-<div>
-    <label asp-for="UserId"></label>
-        <select asp-for="UserId" asp-items="@(new SelectList(MockData.Users, "Id", "FullName"))">
-            <option>Please select one</option>
-        </select>
-</div>
\ No newline at end of file
+<section class="grid grid-cols-1 sm:grid-cols-2">
```
Cybotive, line 3: Slightly different than what was asked for.

SecretSanta/src/SecretSanta.Web/Views/Gifts/Index.cshtml
```diff
@@ -1,11 +1,16 @@
+@{ViewBag.Title = "Gifts";}
+
 @model List<GiftViewModel>
 
 @{int counter = 0;}
-
+<section class="grid grid-cols-1 sm:grid-cols-3">
 @foreach(var groupViewModel in Model) {
```
Cybotive, line 7: Stokes' copy-pasta error. Not functionally catastrophic.

Cybotive (commented):
- The header should have the following elements
  - The header should have a background color and should extend the full width of the screen: ✔
  - The navigation elements should be layed out horizontally instead of vertically: ✔
  - The items should not look like a bulleted list, but should look more like a typical navigation bar: ✔
  - Instead of "Home" the text for that link should be Secret Santa and the font for this should be a bit larger than the other navigation elements: ✔
  - The navigation should be responsive, which means on a small screen the navigation should be hidden until the user clicks on some type of indicator to expand out the navigation elements: ❌ **(The id isn’t linked up to the header but the js is there)**
- The page content should have margin around the whole thing so that the text isn't butted up right against the edge of the screen: ✔
- The user "home" page should
  - When on a small screen, display all users in a single column (First and Last name): ✔
  - When on a medium or larger screen, display users in rows of no more than 3 users per row: ✔
  - Clicking the user on either screen should take the person to an edit screen: ✔
  - There should be an "icon" of some sort on each "tile" that when clicked will delete the user: ❌ **(There is an ‘X’ character instead of icon)**
    - When deleting, the person should be prompted with a confirmation dialog before actually performing the delete: ✔
- The gift "home" page should
  - The gifts should be ordered by priority, where priority 1 items are at the top: ❌ **(Not ordered by priority; Drone (p2) is in front of Rat Poison (p1))**
  - When on a small screen, display all gifts in a single column (Title only): ✔
  - When on a medium or larger screen, display gifts in rows of no more than 3 gifts per row (Title only): ✔
  - Clicking the gift on either screen should take the person to an edit screen: ✔
  - There should be an "icon" of some sort on each "tile" that when clicked will delete the gift: ❌ **(There is an ‘X’ character instead of icon)**
    - When deleting, the person should be prompted with a confirmation dialog before actually performing the delete: ✔
- The user "edit" page should
  - When on a small screen, fields should be lined up in a single column: ✔
  - When on a medium or larger screen, the first name and last name editing should be on the same line (side by side): ✔
  - If there is a data issue with any of the text boxes upon submit, then when the screen is presented again, the input box should reflect an error has occurred: ✔
  - The "save" and "cancel" buttons should have some type of hover effect so the person knows when their cursor is over the button: ✔
- The gift "edit" page should
  - When on a small screen, fields should be lined up in a single column: ✔
  - When on a medium or larger screen, the title and description should be vertically aligned, but the url and priority should be layed out side by side: ❌ **(Title and Description still side by side on “greater than small” screen)**
  - If there is a data issue with any of the text boxes upon submit, then when the screen is presented again, the input box should reflect an error has occurred: ✔
  - The "save" and "cancel" buttons should have some type of hover effect so the person knows when their cursor is over the button: ✔
  - Make it so that the Url and Description fields are not required: ✔
- Fix the title on every page to reflect this is a Secret Santa app instead of a UserGroup app: ❌ **(All pages read “UserGroup – \<page name\>“)**

breaman:
Gifts were not ordered by priority, The "Edit" text should not have been needed, that was suppose to be the default behavior for just clicking on a User/Gift. Overall, though, nice job with the assignment.

PR #99, opened 2021-04-19
Assignment3.

CalumMackenzie-EWU:
Everything is now working properly I believe.

### DIFF
--- a/SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
+++ b/SecretSanta/src/SecretSanta.Web/Controllers/GiftsController.cs
@@ -8,6 +8,8 @@ namespace SecretSanta.Web.Controllers
     {
         public IActionResult Index()
         {
+            //cal: the sort orders the list by priority so that priority 1 gifts are at the top.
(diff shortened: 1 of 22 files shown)


MeguminXIII:
I will review

MeguminXIII (commented):
The navigation elements should be layed out horizontally instead of vertically: ✔
The user "home" page should
The gift "home" page should
The gifts should be ordered by priority, where priority 1 items are at the top: ✔
The user "edit" page should
When on a medium or larger screen, the first name and last name editing should be on the same line (side by side): ✔
A suggestion would be to push them closer together maybe? on my screen they are on opposite sides of my screen with a large amount of white space between them. just an opinion though.

The gift "edit" page should
When on a medium or larger screen, the title and description should be vertically aligned, but the url and priority should be layed out side by side: ✔
Make it so that the Url and Description fields are not required: ✔
Fix the title on every page to reflect this is a Secret Santa app instead of a UserGroup app: ✔

Looks really good, nice job

breaman:
Nice job on the assignment, all requirements were hit. Technically you didn't need the edit text since clicking on the item took you to the edit screen anyway.

PR #106, opened 2021-04-25
Assignment4 with @anthonehi

roylalog000:
code pairing @anthonehi


SecretSanta/test/SecretSanta.Api.Tests/Controllers/UsersControllerTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            public ICollection<User> List()
+            {
+                throw new System.NotImplementedException();
+            }
+
+            // ???
+            public bool Remove(int id)
+            {
+                UserManager manager = new();
+                User? usercheck = manager.GetItem(id);
```
Keboo, line 139: This works because of how the real implementation is setup with a list, but in general when you are implementing a test double for something you would leave yourself a hook to control the behavior of the method without relying on calling the real implementation.

So perhaps similar to what you did with `GetItem` above, simply give yourself a flag to use to to control the return value and have this method simply return that without doing any real work.

Keboo (commented):
Really clean code. Nice use of data driven tests as well. 

- Create a UserController in the API projects
  - Implement endpoints for:
    - Retrieving all users ✔
    - Retriving a single user by id ✔
    - Creating a user ✔
    - Updating a user ✔
    - Deleting a user ✔
  - The UserController should function. ✔
  - Include a `.rest` file for interacting with the controller. ✔
- The `IUserRepository` interface is provided for you. You will need to provide an implementation of this interface within the SecretSanta.Business project. This implementation can simply store the user within a list, dictionary, array, or similar. ✔
- All test projects should be within a test directory ✔
- Create a unit test project for the SecretSanta.Business project. ✔
  - Unit test all public methods ✔
  - Ensure the project is named with a ".Tests" suffix ✔
  - Ensure test class is named appropriately for the class being tested: ✔
  - Name all tests in the form `<MethodBeingTested>_<ConditionBeingTested>_<ExpectedOutcome>()`. You can add  `<NoWarn>CA1707;INTL0003</NoWarn>` to remove the naming warnings. ✔
- Create a Unit Test project for the SecretSanta.Api project. ✔
  - Test all controller public methods. It is up to you to determine the best way to provide the controller's dependencies. ✔
  - Ensure the project is named with a ".Tests" suffix ✔

PR #121, opened 2021-05-02
Assignment 5

ybaghlaf:
just starting the assignment


SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -53,6 +61,8 @@ public ActionResult Delete(int id)
         }
 
         [HttpPut("{id}")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
```
Keboo, line 65: This endpoint also returns NoFound (404) in addition to these two.

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -38,18 +60,18 @@ public IActionResult Edit(UserViewModel viewModel)
         {
             if (ModelState.IsValid)
             {
-                MockData.Users[viewModel.Id] = viewModel;
+                //MockData.Users[viewModel.Id] = viewModel;
```
Keboo, line 63: This should also be invoking the IUserClient Post endpoint to update the the user.

SecretSanta/tests/SecretSanta.Web.Tests/Controllers/UsersControllerTest.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SecretSanta.Web.Api;
+using SecretSanta.Web.Tests.Api;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Tests
+{
+    [TestClass]
+    public class UsersControllerTest
```
Keboo, line 13: This is a good start, but this only tests two of the end points. The Delete and Edit endpoints should also be tested.

SecretSanta/tests/SecretSanta.Web.Tests/SecretSanta.Web.Tests.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
Keboo, line 1: This is close, but it is in a "tests" namespace, but the target directory name should be "test" not "tests"

Keboo (commented):
  - Add a new Dto namespace ❌
  - Convert the `UserController` from using the `MockData` to using the client to get data from the API service. ❌ - Edit does not post using service
  - All endpoints should be tested. ❌ - only index and create are tested
  - It should be inside of the test directory. ✔ - It was named "tests" not "test"
  - It should have functional tests for all endpoints in the `UserController` that access the API. ❌ - Only Index and Create are tested

ybaghlaf:
I think I missed that one Kevin! Whoops

PR #122, opened 2021-05-03
Assignment5

Cybotive:


SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -19,20 +19,37 @@ public UsersController(IUserRepository repository)
         }
 
         [HttpGet]
-        public IEnumerable<User> Get()
+        [ProducesResponseType(typeof(IEnumerable<User>), StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        public IEnumerable<FullUser> Get()
```
Fervidusletum, line 24: Line 22 doesn't match the return type of the function (`user`, should be `fulluser`)

Fervidusletum (changes requested):
- - Add a new Dto namespace ✔
- - The UserController should only accept or return classes from the Dto namespace. Create any needed classes there and update the controller to use them. You will need to convert between the DTO classes and the relevant entities as needed. ✔
- - Add OpenAPI (swagger) endpoint ✔
- - Use the OpenAPI to generate a C# client with relevant interfaces ✔
- - Decorate all relevant API endpoints in the UserController with the appropriate ProducesResponseType ✔
- - Register the generated UserClients class using its interface with the Dependency Injection (DI) service. ✔
- - Update the UserController to take in the generated client as a dependency. ✔
- - Convert the UserController from using the MockData to using the client to get data from the API service. ✔
- - Update the controller to convert back and forth between the ViewModel classes and the API DTO classes as needed. ✔
- - The controller tests should be interation tests not unit tests. ✔
- - All endpoints should be tested. ❌
- - The UserController should use a test double for the IUserRepository to avoid needing to retest business and data logic. ✔
- - It should be inside of the test directory. ✔
- - It should have nullable reference types enabled. ✔
- - It should have functional tests for all endpoints in the UserController that access the API. ❌
- - It should use a test double for the API service. This test double should be used to validate the controller's interaction with the API without actually calling the API. ✔
- Extra Credit:
- - Rather than coping data between view models and DTOs, and DTOs and entities; use AutoMapper to handle it for you. ❌


Looks good for the most part!  Just some more endpoint testing in both api and web userscontroller to do.

Fervidusletum:
Found a new issue: when running the client + server, clicking on a user to edit will generate an exception.
`InvalidOperationException: The model item passed into the ViewDataDictionary is of type 'System.Runtime.CompilerServices.AsyncTaskMethodBuilder1+AsyncStateMachineBox1[SecretSanta.Web.Api.FullUser,SecretSanta.Web.Api.UsersClient+<GetAsync>d__15]', but this ViewDataDictionary instance requires a model item of type 'SecretSanta.Web.ViewModels.UserViewModel'.`

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
Cybotive: I didn't know exactly what the server would respond with so I changed it back to User but you're likely right.

Cybotive:
> 
> 
> Found a new issue: when running the client + server, clicking on a user to edit will generate an exception.
> `InvalidOperationException: The model item passed into the ViewDataDictionary is of type 'System.Runtime.CompilerServices.AsyncTaskMethodBuilder1+AsyncStateMachineBox1[SecretSanta.Web.Api.FullUser,SecretSanta.Web.Api.UsersClient+<GetAsync>d__15]', but this ViewDataDictionary instance requires a model item of type 'SecretSanta.Web.ViewModels.UserViewModel'.`

Did you run `npm run build:dev` first? Could be a valid error but I suspect that would fix it.

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -19,41 +19,68 @@ public UsersController(IUserRepository repository)
...
             if (user is null) return NotFound();
-            return user;
+
+            return UserToFullUser(user);
         }
 
         [HttpDelete("{id}")]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)] // Using 204 instead of 200 since this method doesn't return any content
+        [ProducesResponseType(StatusCodes.Status200OK)] // Doesn't happen in this implementation
```
Keboo, line 53: Since it doesn't happen it can be omitted.
```suggestion
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/UserControllerTests.cs
```diff
@@ -5,16 +5,317 @@
...
+        [TestMethod]
+        public void UsersController_WithValidParameter_DoesNotThrowException()
+        {
+            //Arrange - Nothing to arrange
+
+            //Act
+            Api.Controllers.UsersController controllerTemp = new(new TestableUserRepository());
+
+            //Assert - Nothing to assert
+        }
```
Keboo, line 45: In general you don't need to test this case as it will be covered by your other tests.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/UserControllerTests.cs
```diff
@@ -5,16 +5,317 @@
...
+            List<User>? userList = await response.Content.ReadFromJsonAsync<List<User>?>();
+            
+            //Assert
+            response.EnsureSuccessStatusCode();
+
+            if(userList is null)
+            {
+                Assert.Fail("No List was received. \'" + nameof(userList) + "\' is null.");
+                return;
+            }
```
Keboo, line 70: Typically you want to avoid any conditional statements in your tests like this. A better way to handle this case would be to simply assert that the `userList` is not null.
```suggestion
            Assert.IsNotNull(userList);
```

Keboo (commented):
  - Add a new Dto namespace ✔

PR #123, opened 2021-05-03
M johnson assignment5

thecactus1:
Not sure why automerge is not working but idk


Keboo:
It appears you have two branches that only differ by casing. This subtle issue appears to be causing quite a bit of trouble with various git tools, including GitHub. Would you mind removing one of them? You can see it [here](https://github.com/thecactus1/EWU-CSCD379-2021-Spring/branches/all). Based on the timestamps it looks like the one with the capital J is the one that you want to keep.

thecactus1:
Thanks for bringing that to my attention. Git can be a real brat sometimes, especially since we don’t really get taught this kind of usage explicitly.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

thecactus1:
I have a new test files on my computer do you mind if I update that tonight

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

Keboo:
Go ahead, and send me an e-mail when it is done.

PR #126, opened 2021-05-03
Assignment5

jmplett:


Cybotive:
I will review this

Cybotive (commented):
﻿In this assignment we are going integrate the Web project with the API project, and fix some testing failures from the last assignment.

# Assignment
  - Add a new Dto namespace ✔
  - Register the generated `UserClients` class using its interface with the Dependency Injection (DI) service. ❌
  - Update the `UserController` to take in the generated client as a dependency. ❌
  - Convert the `UserController` from using the `MockData` to using the client to get data from the API service. ❌
  - Update the controller to convert back and forth between the ViewModel classes and the API DTO classes as needed. ❌
  - The controller tests should be [interation tests](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-5.0) not unit tests. ✔❌ **(Not enough code to determine)**
  - All endpoints should be tested. ❌
  - It should be inside of the test directory. ❌
  - It should have nullable reference types enabled. ❌

Extra Credit:
- Rather than coping data between view models and DTOs, and DTOs and entities; use [AutoMapper](https://automapper.org/) to handle it for you. ❌ **(Not attempted)**

**Explanations for ❌'s are in code comments as specified in Homework.md. ❌'s without explanation are due to having no related changes.**

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -19,12 +20,16 @@ public UsersController(IUserRepository repository)
...
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
         public IEnumerable<User> Get()
         {
             return Repository.List();
         }
 
         [HttpGet("{id}")]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [ProducesResponseType(typeof(User), StatusCodes.Status200OK)]
         public ActionResult<User?> Get(int id)
```
Cybotive, line 33: ❌ Explanation: Still accepts SecretSanta.Data.User

PR #127, opened 2021-05-03
Assignment 5

MeguminXIII:
As done as it's gonna get. Needs more tests though.


NicholasJohnsonDev:
I'll review this.

NicholasJohnsonDev:
Everything builds. Tests incomplete and not all passed. When I ran API then Web and navigated to users I got an error:

InvalidCastException: Unable to cast object of type 'System.Collections.Generic.List`1[SecretSanta.Web.Api.User]' to type 'System.Collections.Generic.ICollection`1[SecretSanta.Web.Api.DtoUser]'.
SecretSanta.Web.Controllers.UsersController.Index() in UsersController.cs, line 21
 21.           ICollection<DtoUser?> users = (ICollection<DtoUser?>)await UserClient.GetAllAsync();


Update the API project with the following:
Add a new Dto namespace ✔

Update the Web project with the following:

The controller tests should be interation tests not unit tests. ❌
All endpoints should be tested. ❌
The UserController should use a test double for the IUserRepository to avoid needing to retest business and data logic. ❌

Create a new SecretSanta.Web.Tests projecs with the following:
It should be inside of the test directory. ❌
It should have nullable reference types enabled. ❌
It should have functional tests for all endpoints in the UserController that access the API. ❌
It should use a test double for the API service. This test double should be used to validate the controller's interaction with the API without actually calling the API. ❌

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -15,24 +18,39 @@ public class UsersController : ControllerBase
 
         public UsersController(IUserRepository repository)
         {
-            Repository = repository ?? throw new System.ArgumentNullException(nameof(repository));
+            Repository = repository ?? throw new ArgumentNullException(nameof(repository));
         }
 
         [HttpGet]
-        public IEnumerable<User> Get()
+        [ProducesResponseType(typeof(IEnumerable<User>), StatusCodes.Status200OK)]
```
Keboo, line 25: This is indicating the wrong response type. 
```suggestion
        [ProducesResponseType(typeof(IEnumerable<DtoUser>), StatusCodes.Status200OK)]
```

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -43,17 +61,35 @@ public ActionResult Delete(int id)
...
-            if (user is null)
+            if (dtoUser is null)
             {
                 return BadRequest();
             }
-            return Repository.Create(user);
+            int id;
+            if(Repository.List().Count == 0){
+                id = 0;
+            }
```
Keboo, line 75: nit: We typically place braces on their own lines.
```suggestion
            if(Repository.List().Count == 0)
            {
                id = 0;
            }
```

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -43,17 +61,35 @@ public ActionResult Delete(int id)
...
+            return dtoUser;
         }
 
         [HttpPut("{id}")]
-        public ActionResult Put(int id, [FromBody] User? user)
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        public ActionResult Put(int id, [FromBody] DtoUser? user)
```
Keboo, line 92: This works, but it takes in a full `DtoUser` which has more properties on it then the update will actually use (since it only leverages FirstName/LastName).

SecretSanta/test/SecretSanta.Api.Tests/WebApplicationFactory.cs
```diff
@@ -0,0 +1,19 @@
...
+using SecretSanta.Api.Tests.Business;
+using Microsoft.Extensions.DependencyInjection;
+using SecretSanta.Business;
+
+
+namespace SecretSanta.Api.Tests
+{
+    public class WebApplicationFactory : WebApplicationFactory<Startup>
+    {
+        public TestableUserRepository Tur {get;} = new();
```
Keboo, line 12: nit: We generally avoid using acronyms unless they are very well known ones.

Keboo (commented):
  - Add a new Dto namespace ✔

PR #128, opened 2021-05-03
Assignment 5

ZStuefen:


delta-plus:
I will review this.

delta-plus (commented):
Looks pretty good! Only missing a few small pieces.

# Assignment
  - Add a new Dto namespace ✔
  - The `UserController` should only accept or return classes from the Dto namespace. Create any needed classes there and update the controller to use them. You will need to convert between the DTO classes and the relevant entities as needed. ❌ (Post method both accepts and returns a User object. Put is good, though.)
  - Decorate all relevant API endpoints in the `UserController` with the appropriate [ProducesResponseType](https://docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-5.0#iactionresult-type) ❌ (Only Delete and Post are decorated.)
  - Register the generated `UserClients` class using its interface with the Dependency Injection (DI) service. ✔ (The correct code is there in Startup.cs but it's been commented out for some reason.)
  - Convert the `UserController` from using the `MockData` to using the client to get data from the API service. ✔ (The Edit method is still updating directly via MockData.Users[Id] instead of PutAsync, but I guess that's not getting data, it's setting it.)
  - All endpoints should be tested. ❌ (Post and Delete not tested.)
  - The `UserController` should use a test double for the `IUserRepository` to avoid needing to retest business and data logic. ❌ (Does use a test double, except for the Get_WithData_ReturnsUsers method.)
  - It should have functional tests for all endpoints in the `UserController` that access the API. ❌ (Missing tests for Edit and Delete.)

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -27,14 +28,24 @@ public IEnumerable<User> Get()
         [HttpGet("{id}")]
         public ActionResult<User?> Get(int id)
```
Keboo, line 29: Both Get methods should be returning DTO objects.

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -43,6 +54,8 @@ public ActionResult Delete(int id)
         }
 
         [HttpPost]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(typeof(User), StatusCodes.Status200OK)]
         public ActionResult<User?> Post([FromBody] User? user)
```
Keboo, line 59: This method should be taking in a DTO object as a parameter.

SecretSanta/src/SecretSanta.Web/Startup.cs
```diff
@@ -7,15 +7,23 @@
...
+        public static System.Net.Http.HttpClient ApiClient = new()
+        {
+            BaseAddress = new Uri("https://localhost:5101")
+        };
+
         // This method gets called by the runtime. Use this method to add services to the container.
         // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
+            //services.AddScoped<IUsersClient, UsersClient>(_ => new UsersClient(ApiClient));
```
Keboo, line 25: I am not sure why this is commented out. It looks to be the correct line of code.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/UserControllerTests.cs
```diff
@@ -1,20 +1,117 @@
...
+            {
+                new UsersController(null!);
+            }
+            catch(ArgumentNullException e)
+            {
+                Assert.AreEqual("userRepository", e.ParamName);
+                return;
+            }
+            Assert.Fail("No exception thrown");
+        }
```
Keboo, line 39: This shows three possible ways to check that the constructor throws the appropriate argument null exception. The expectation is that you pick one of methods, and remove the others.

SecretSanta/test/SecretSanta.Web.Tests/Controllers/UsersControllerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SecretSanta.Web.Api;
+using SecretSanta.Web.Tests.Api;
+using SecretSanta.Web.ViewModels;
+
+namespace SecretSanta.Web.Tests
+{
+    [TestClass]
+    public class UsersControllerTests
```
Keboo, line 13: This is a good start, but it only tests the Index and Create endpoints. The Delete and Edit endpoints are not tested.

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -1,14 +1,36 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
+using SecretSanta.Web.Api;
 using SecretSanta.Web.Data;
 using SecretSanta.Web.ViewModels;
 
 namespace SecretSanta.Web.Controllers
 {
     public class UsersController : Controller
```
Keboo, line 11: This is close, but the Edit endpoint is still using the MockData.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/UserControllerTests.cs
```diff
@@ -1,20 +1,117 @@
...
+
+            //Act
+            ActionResult<User?> result = controller.Get(-1);
+
+            //Assert
+            Assert.IsTrue(result.Result is NotFoundResult);
+        }
+
+        [TestMethod]
+        public async Task Put_WithValidData_UpdatesUser()
```
Keboo, line 90: This is a start, but the expectation was that the API tests be converted to integration tests.

PR #132, opened 2021-05-03
Assignment5 with @anthonehi

roylalog000:
code pairing @anthonehi


jmplett:
I will review

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -43,33 +53,41 @@ public ActionResult Delete(int id)
...
                 return BadRequest();
             }
-            return Repository.Create(user);
+            return Repository.Create(myUser);
         }
 
+
         [HttpPut("{id}")]
-        public ActionResult Put(int id, [FromBody] User? user)
+        public ActionResult Put(int id, [FromBody]UpdatedUser? updatedEvent)
```
jmplett, line 70: Missing status codes

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -1,36 +1,62 @@
...
             return View(viewModel);
         }
 
+
+
+
         public IActionResult Edit(int id)
         {
-            return View(MockData.Users[id]);
+            return View(MockData.Users[id - 1]);
```
jmplett, line 59: Needs to be implemented with the DTO

SecretSanta/test/SecretSanta.Api.Tests/Controllers/UserControllerTests.cs
```diff
@@ -15,6 +15,6 @@ namespace SecretSanta.Api.Tests.Controllers
     [TestClass]
     public class UserControllerTests
     {
-        
+
```
jmplett, line 18: Tests are not done

jmplett:
  - Add a new Dto namespace ✔
  - Decorate all relevant API endpoints in the `UserController` with the appropriate [ProducesResponseType](https://docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-5.0#iactionresult-type) ❌
  - Register the generated `UserClients` class using its interface with the Dependency Injection (DI) service. ❌
  - Update the `UserController` to take in the generated client as a dependency. ❌
  - Convert the `UserController` from using the `MockData` to using the client to get data from the API service. ❌
  - Update the controller to convert back and forth between the ViewModel classes and the API DTO classes as needed. ❌
  - The controller tests should be [interation tests](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-5.0) not unit tests. ❌
  - All endpoints should be tested. ❌

Good start just have some loose ends to fix and them complete the rest of the assignment.

Keboo:
@jmplett though perfectly fine to review, when possible see if there is a PR that was not done as a [pair](https://github.com/IntelliTect-Samples/EWU-CSCD379-2021-Spring/blob/main/Docs/Homework.md#pairing) when looking to review :).

PR #134, opened 2021-05-03
Assignment 5

Fervidusletum:
Finished


MeguminXIII:
I will Review

SecretSanta/test/SecretSanta.Web.Tests/Controllers/UsersControllerTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            HttpClient client = Factory.CreateClient();
+
+            HttpResponseMessage response = await client.GetAsync("/users/");
+
+            response.EnsureSuccessStatusCode();
+            Assert.AreEqual(1, usersClient.GetAllAsyncInvokeCount);
+        }
+
+        [TestMethod]
+        public void Index_RetreivesValidDtoList_CreatesValidViewModelList()
```
MeguminXIII, line 41: Everything below this needs implementing

SecretSanta/test/SecretSanta.Api.Tests/Controllers/UsersControllerTests.cs
```diff
@@ -0,0 +1,188 @@
...
+        public void Put_GivenInvalidUser_Returns400()
+        {
+            Assert.Fail();
+        }
+
+        [TestMethod]
+        public void Put_GivenInvalidId_Returns404()
+        {
+            Assert.Fail();
+        }
```
MeguminXIII, line 156: needs implementing

MeguminXIII (commented):
Assignment
Update the API project with the following
Add a new Dto namespace ✔
Update the Web project with the following
The controller tests should be interation tests not unit tests. ✔
All endpoints should be tested. ❌
some still not implemented, will put in a different comment

The UserController should use a test double for the IUserRepository to avoid needing to retest business and data logic. ✔
Create a new SecretSanta.Web.Tests project with the following:
It should be inside of the test directory. ✔
It should have nullable reference types enabled. ✔
It should have functional tests for all endpoints in the UserController that access the API. ❌
Most aren't implemented yet, will put in a separate comment

It should use a test double for the API service. This test double should be used to validate the controller's interaction with the API without actually calling the API. ✔
Extra Credit:

Rather than coping data between view models and DTOs, and DTOs and entities; use AutoMapper to handle it for you.❌

Looks good, just need to finish the tests

SecretSanta/src/SecretSanta.Api/Dto/UserDtoFnLn.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SecretSanta.Api.Dto
+{
+    public class UserDtoFnLn
```
Keboo, line 3: nit: We typically avoid abbreviations unless they are common and very well known.

Keboo (commented):
  - Add a new Dto namespace ✔
  - All endpoints should be tested.✔
  - The `UserController` should use a test double for the `IUserRepository` to avoid needing to retest business and data logic.✔

PR #135, opened 2021-05-03
Assignment5 unfinished

CalumMackenzie-EWU:


SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -19,22 +21,48 @@ public UsersController(IUserRepository repository)
         }
 
         [HttpGet]
-        public IEnumerable<User> Get()
+        [ProducesResponseType(typeof(IEnumerable<User>), StatusCodes.Status200OK)]
+        //public IEnumerable<User> Get()
+        public IEnumerable<FullUser> Get()
```
Keboo, line 26: There is a disjoint here between the return type (`IEnumerable<FullUser>`) and the declared returned type (`IEnumerable<User>`).
```suggestion
        [ProducesResponseType(typeof(IEnumerable<FullUser>), StatusCodes.Status200OK)]
        //public IEnumerable<User> Get()
        public IEnumerable<FullUser> Get()
```

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -43,17 +71,30 @@ public ActionResult Delete(int id)
...
+
+            return user;
         }
 
         [HttpPut("{id}")]
-        public ActionResult Put(int id, [FromBody] User? user)
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        public ActionResult Put(int id, [FromBody] FullUser? user)
```
Keboo, line 97: In this case it looks like you forgot to change the parameter to `UpdateUser`
```suggestion
        public ActionResult Put(int id, [FromBody] UpdateUser? user)
```

Keboo (commented):
The implementation looks pretty good, only a couple things in your API controller. Just missing the tests.

  - Add a new Dto namespace ✔
  - The controller tests should be [integration tests](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-5.0) not unit tests. ❌
  - All endpoints should be tested.❌
  - The `UserController` should use a test double for the `IUserRepository` to avoid needing to retest business and data logic.❌
  - It should be inside of the test directory. ❌
  - It should have nullable reference types enabled. ❌

PR #136, opened 2021-05-03
Partially completed Assignment 5.

delta-plus:
Still need to write some more tests for Api and Web. Had a heck of time trying to parse the JSON responses. Parse errors are the worst. Check out my Api test file if you want to see my pain and suffering.


ZStuefen:
I will review this

ZStuefen:
  - Add a new Dto namespace ✔
      - Missing delete test

SecretSanta/test/SecretSanta.Web.Tests/Controllers/UsersControllerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+//    Not sure how you'd test this because I can't really find an endpoint to post to 
+//    (the Delete button simply uses asp-action="Delete")
+//    and furthermore, the Delete button just redirects to "api/Users/[id]" instead of 
+//    actually deleting ever since implementing the generated UsersClient. Weird.
+
+//    HttpResponseMessage response = await client.PostAsync(???, ???);
+
+//    response.EnsureSuccessStatusCode();
+//    Assert.AreEqual(1, usersClient.DeleteAsyncInvocationCount);
```
Keboo, line 92: You are very close. The delete does in fact post back to the controller. So you would invoke it with a Post call.
You also can test it by verifying that when you post to that endpoint that it does in fact call the Delete endpoint on the API service. You can see an example [here](https://github.com/IntelliTect-Samples/EWU-CSCD379-2021-Spring/blob/Assignment5Complete/SecretSanta/test/SecretSanta.Web.Tests/Controllers/UsersControllersTests.cs#L164-L187)

Keboo (commented):
  - Add a new Dto namespace ✔
  - All endpoints should be tested.✔
  - The `UserController` should use a test double for the `IUserRepository` to avoid needing to retest business and data logic.✔

PR #163, opened 2021-05-16
Assignment7

worseTyler:


ZStuefen:
I will review this

ZStuefen:
# Assignment

.github/workflows/AssignmentBuild.yml
```diff
@@ -28,6 +28,9 @@ jobs:
         npm install
         npm run build:prod
 
+    - name: Trust Certificates
```
breaman, line 31: Nice, going to have to make sure I do this myself.

SecretSanta/src/SecretSanta.Web/Views/Users/Index.cshtml
```diff
@@ -1,18 +1,13 @@
-@model List<UserViewModel>
-
-
-<section class="flex flex-wrap mb-6">
-    @foreach(var userViewModel in Model) {
+<section class="flex flex-wrap mb-6" x-data="SecretSanta.setupUsers()" x-init="mounted()">
+    <template x-for="(currentUser, index) in users" :key="index">
         <section class="w-full sm:w-1/3">
-            <a asp-action="Edit" asp-route-id=@userViewModel.Id>
+            <a :href="'https://localhost:5001/users/edit/' + currentUser.id">
```
breaman, line 4: Not sure why you hardcoded the url to be port 5001 here.

SecretSanta/test/SecretSanta.Web.EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        [TestMethod]
+        public async Task NavigateBetweenPages()
+        {
+            var page = await Browser.NewPageAsync();
+            var response = await page.GoToAsync(LocalHost);
+
+            Assert.IsTrue(response.Ok);
+
+            await page.ClickAsync("text=Users");
+            Assert.AreEqual<string>(LocalHost + "Users", page.Url);
```
breaman, line 55: I like this way of validating that each page loaded, nice job.

breaman (commented):
Nice work overall on the assignment.

PR #165, opened 2021-05-17
Assignment 7

ZStuefen:
Completed Assignment but could not check E2E tests because of the net::ERR_CERT_AUTHORITY_INVALID error.


delta-plus:
I will review this.

delta-plus (commented):
Nice work. A few things could be improved but not bad overall.

  - Create a Gift (should assert that the number of gifts after the creation is one greater than the number of gifts before the user was created) ❌ (Test doesn't work. It throws a "waiting for selector" error because the test code has 'await page.TypeAsync("input#UserId", "1");' instead of 'await page.SelectOptionAsync("select#UserId", "1");')
  - Modify the last gift in the list (should assert that the updated data was successfully saved/displayed to the screen) ❌ (Fails with "Assert.AreEqual failed. Expected:<UpdatedGift>. Actual:<UpdatedGiftDrone>." This means you either need to clear the title field before typing into it, use .IsTrue(text.Contains("UpdatedGift")) instead of .AreEqual(), or both.)

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            var gifts = await page.QuerySelectorAllAsync("body > section > section > section");
+            int giftNum = gifts.Count();
+
+            await page.ClickAsync("text=Create");
+
+            await page.TypeAsync("input#Title", "Gift");
+            await page.TypeAsync("input#Url", "google.com");
+            await page.TypeAsync("input#Description", "A new gift");
+            await page.TypeAsync("input#Priority", "1");
+            await page.TypeAsync("input#UserId", "1");
```
breaman, line 126: This test failed because you didn't use the SelectOptionAsync

breaman (commented):
Overall not bad, but you also committed way too many files which made digging through things a little challenging. You might want to verify that the "Files Changed" seems reasonable (406 was quite a few files).

PR #167, opened 2021-05-17
Assignment7

OkayNiceBrian:
I spent so long trying to troubleshoot why the typescript just isn't working with my User, I have no idea why. If a kind code reviewer could help me out, I'd appreciate it greatly!


SecretSanta/src/SecretSanta.Api/Startup.cs
```diff
@@ -25,10 +41,12 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 app.UseDeveloperExceptionPage();
             }
 
+            app.UseRouting();
+            
             app.UseOpenApi();
             app.UseSwaggerUi3();
 
-            app.UseRouting();
+            app.UseCors();
```
Fervidusletum, line 49: It looks like you removed `app.UseRouting();` here, try adding that back in after `app.UseCors();` and see if that helps.

breaman (commented):
Great job on the assignment, you hit all the elements that were asked for.

PR #169, opened 2021-05-17
Completed Assignment 7.

delta-plus:
Voilà!


Fervidusletum:
## Assignment

## Comments
- E2E tests were not added to sln.
- Otherwise, looks good! 👍

delta-plus:
Thanks for the tip on adding the test project to the .sln. Done!

breaman (commented):
Excellent job on the assignment. I like how you went back and updated based on PR feedback as well.

PR #170, opened 2021-05-17
Assignment7

jmplett:


worseTyler:
I will review this

SecretSanta/test/SecretSanta.EndToEnd.Tests/EndToEndTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            await Task.WhenAll(
+                Task.Run(
+                    async () => { response = await page.WaitForNavigationAsync(); }), 
+                page.ClickAsync("text=Groups"));
+
+            Assert.IsTrue(response.Ok);
+        }
+
+        [TestMethod]
+        public async Task NavigateToGifts()
```
worseTyler, line 89: I feel this can be grouped into one test instead of three but that's my opinion

worseTyler:
Create a Gift (should assert that the number of gifts after the creation is one greater than the number of gifts before the user was created) ❌
Modify the last gift in the list (should assert that the updated data was successfully saved/displayed to the screen) ❌
Delete the last gift in the list (should assert that the number of gifts after the delete is one less than the number of gifts before the deletion) ❌

While you have things in your typescript, you didn't actually change any of the Gifts view pages to actually use the typscript. Instead they still use all of the asp.net.
You also only test like half of everything that you are supposed to test

worseTyler:
image
I also ran your E2E tests and they all gave me this error

breaman (commented):
Nice job on the assignment. I see that you ended up generating the typescript client but then never used it. Since I never covered that in class, not all that concerned, just saw you made the effort and then didn't use it.

PR #171, opened 2021-05-17
Assignment 7

MeguminXIII:


jamesgreiner:
I will review this.

jamesgreiner (commented):
Verify that when landing on the homepage, Secret Santa is displayed in the navigation banner ✔❌
Verify that all 3 pages (Users, Groups, Gifts) can be reached via clicking on the navigation elements ✔❌
Create a Gift (should assert that the number of gifts after the creation is one greater than the number of gifts before the user was created) ✔❌
Modify the last gift in the list (should assert that the updated data was successfully saved/displayed to the screen) ✔❌
Delete the last gift in the list (should assert that the number of gifts after the delete is one less than the number of gifts before the deletion) ✔❌
All of these tests should be able to be executed without having to manually start the Web and Api projects ✔❌


Looks good so far. The AplineJS stuff appears to be there and like you said, it should just be the tests at this point. Keep up the good work!

breaman (commented):
Unfortunately none of your code actually ran due to the fact that the generated ts file was not checked in and it was also not auto-generated in the api project like the cs file did.

PR #172, opened 2021-05-17
Assignment7 with @anthonehi

roylalog000:
code pairing


OkayNiceBrian:
I'll review this.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
...
+            var response = await page.GoToAsync(localhost);
+
+            Assert.IsTrue(response.Ok);
+
+            var headerContent = await page.GetTextContentAsync("body > header > div > a");
+            Assert.AreEqual("Secret Santa", headerContent);
+        }
+
+        [TestMethod]
+        public async Task CreateUser()
```
OkayNiceBrian, line 39: Should be testing for Gift (not User) in all cases. It's understandable because the Assignment7 readme went through some changes.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            Assert.IsTrue(response.Ok);
+
+            await page.ClickAsync("text=Users");
+
+            await page.ScreenshotAsync("create.png");
+
+            // make sure we have 5 speakers here
+            var users = await page.QuerySelectorAllAsync("body > section > section > section");
+            Assert.AreEqual(5,users.Count());
```
OkayNiceBrian, line 59: When I ran the Unit test, it said that it returned 0 users instead of the expected 5. I suspect this is happening because of the delay between reaching the user page and when your users are loaded from the api (if you load this in your site you'll notice a small delay). I bet this could be fixed by making your test wait a second before counting the users (although maybe there's a better way).

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
...
+            var page = await browser.NewPageAsync();
+            var response = await page.GoToAsync(localhost);
+
+            Assert.IsTrue(response.Ok);
+
+            await page.ClickAsync("text=Users");
+
+            var sectionText = await page.GetTextContentAsync("body > section > section > section:last-child");
+
+            Assert.IsTrue(sectionText.Contains("Miracle Max"));
```
OkayNiceBrian, line 117: This line fails the unit test. I'm not sure if that syntax works (maybe it does, I just don't know). It is probably better to use Assert.AreEqual("Miracle Max", sectionText);

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
...
+            page.Dialog += (_, args) => args.Dialog.AcceptAsync();
+
+            await page.ClickAsync("body > section > section > section:last-child > a > section > button");
+
+            var users = await page.QuerySelectorAllAsync("body > section > section > section");
+            Assert.AreEqual(5, users.Count());
+        }
+
+        [TestMethod]
+        public async Task EditUser()
```
OkayNiceBrian, line 99: This test currently doesn't edit a user, just checks the value of the last user in the list.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using PlaywrightSharp;
+using System.Linq;
+
+namespace SecretSanta.Web.Tests
+{
+    [TestClass]
+    public class EndToEndTests
```
OkayNiceBrian, line 9: You don't verify that you can navigate to all three views (User, Gift, and Group).

OkayNiceBrian (commented):
Verify that all 3 pages (Users, Groups, Gifts) can be reached via clicking on the navigation elements ❌
Create a Gift (should assert that the number of gifts after the creation is one greater than the number of gifts before the user was created) ❌
Modify the last gift in the list (should assert that the updated data was successfully saved/displayed to the screen) ❌

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            Assert.IsTrue(response.Ok);
+
+            await page.ClickAsync("text=Users");
+
+            page.Dialog += (_, args) => args.Dialog.AcceptAsync();
+
+            await page.ClickAsync("body > section > section > section:last-child > a > section > button");
+
+            var users = await page.QuerySelectorAllAsync("body > section > section > section");
```
breaman, line 94: Since you ended up testing users instead of gifts (which I said was ok since I did change the assignment) you then needed to make these tests work for alpinejs like we did in class. You are missing this piece of code before the QuerySelectorAllAsync call. Since the call is making a javascript call, you need to wait for the page to load up the response and render the data.
```
await page.WaitForSelectorAsync("body > section > section > section");
```

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
breaman: Like the reviewer mentioned, this isn't testing the editing of a user at all, it is just getting the last user in the list and never changing the data. Unfortunately, this is still failing because like in your earlier test you are not waiting for the the page to be rendered before trying to pull text.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
breaman: The reviewer called this out and still it was not corrected unfortunately.

breaman (commented):
Unfortunately missing some key elements of the testing portion of the assignment as well as the create/edit page had some issues.

PR #174, opened 2021-05-17
Assignment 7

Fervidusletum:


Cybotive:
I will review this.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            IPage? page = await browser.NewPageAsync();
+            IResponse? response = await page.GoToAsync(localhost);
+
+            await page.ClickAsync($"text={indexpage}");
+
+            Assert.IsTrue(page.Url.Contains(indexpage), $"Could not reach {indexpage} index.");
+
+            string? headerContent = await page.GetTextContentAsync($"#headerNav > ul > li:nth-child({nc}) > a");
+            Assert.AreEqual(indexpage, headerContent, $"{indexpage} not found in nav banner.");
```
Cybotive, line 60: I'm not entirely sure what this check is for since every page has these elements. Is this to catch the case where the returned page is blank due to a server error?

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            IPage? page = await browser.NewPageAsync();
+            IResponse? response = await page.GoToAsync(localhost);
+
+            await page.ClickAsync("text=Gifts");
+
+            string? giftText = await page.GetTextContentAsync("body > section > section > section:last-child > a > section > div");
+            string modifiedText = "Modified Gift";
+
+            Assert.AreNotEqual<string>(giftText, modifiedText, "Test trying to update gift with the same text.");
```
Cybotive, line 108: Isolated from the other tests this seems like it would fail if you tried running the test twice in a row.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            IResponse? response = await page.GoToAsync(localhost);
+
+            await page.ClickAsync("text=Gifts");
+
+            int numGifts = (await page.QuerySelectorAllAsync("body > section > section > section")).Count();
+
+            page.Dialog += (_, args) => args.Dialog.AcceptAsync();
+            await page.ClickAsync("body > section > section > section:last-child > a > section > form > button");
+
+            int numGiftsAfterCreate = (await page.QuerySelectorAllAsync("body > section > section > section")).Count();
```
Cybotive, line 139: Copy pasta.
```suggestion
            int numGiftsAfterDelete = (await page.QuerySelectorAllAsync("body > section > section > section")).Count();
```

Cybotive (commented):
**Nice Work!**
The only comment I'd make is that your navigation tests don't directly verify which page you navigated to. However the page Title wasn't updated by them so it would be hard to check (especially when there's no data in the api at the time) so I'd guess it is out of scope for this assignment.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
Fervidusletum: Yep, just making sure the url is correct and the page loaded.  I couldn't figure out how to test for an OK response using a click action.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
Fervidusletum: Good catch, thanks!

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            string? headerContent = await page.GetTextContentAsync("body > header > div > a");
+            Assert.AreEqual("Secret Santa", headerContent, "Secret Santa not found in nav banner.");
+        }
+
+        [DataTestMethod]
+        [DataRow("Users", 1)]
+        [DataRow("Groups", 2)]
+        [DataRow("Gifts", 3)]
+        public async Task NavTo_IndexPages(string indexpage, int nc)
```
breaman, line 43: Nice use of data driven tests here.

breaman (commented):
Nice job on the assignment as well as making corrections based on PR feedback. Great work.

PR #175, opened 2021-05-17
Assignment7

Cybotive:


jmplett:
I will review

jmplett:
  - Verify that when landing on the homepage, Secret Santa is displayed in the navigation banner ❌
  - Verify that all 3 pages (Users, Groups, Gifts) can be reached via clicking on the navigation elements ❌
  - Create a Gift (should assert that the number of gifts after the creation is one greater than the number of gifts before the user was created) ❌
  - Modify the last gift in the list (should assert that the updated data was successfully saved/displayed to the screen) ❌
  - Delete the last gift in the list (should assert that the number of gifts after the delete is one less than the number of gifts before the deletion) ❌
  - All of these tests should be able to be executed without having to manually start the Web and Api projects ❌

It appears that all your tests have failed on my machine. It looks like you know this based off your comments in your code but if you are getting other results I am more than happy to try them again in the case it might be my machine causing the issue in that case.  Also it appears that you only did half of the tests wanted. Just for simplification I just put X's since even though things were started to be implemented they didn't work as intended and are not entirely complete. 

  - Index page should fetch the Users via javascript directly from the API ❌
  - Create page should submit the Users information via javascript directly to the API. Upon successful save, the user should be redirected back to the Users index page ❌
  - Edit page should submit the Users information via javascript directly to the API. Upon successful update, the user should be redirected back to the Users index page ❌
  - Delete operation should submit the request directly to the API via javascript and then refresh the Index page via a javascript call directly to the API ❌

Looks like you have not really implemented this the alpine/javascript portion yet.

So far you have a good start just need to finished up the loose ends on the tests and then also do the javascript portion.

Cybotive:
I've already done first and last item in Playwright section. Just have to write out the rest of the tests for that section

SecretSanta/test/SecretSanta.E2ETests/EndToEndTests.cs
```diff
@@ -0,0 +1,216 @@
...
+            var response = await page.GoToAsync(localhost);
+
+            Assert.IsTrue(response.Ok);
+            
+            await Task.WhenAll(
+                Task.Run(
+                    async () => { response = await page.WaitForNavigationAsync(); }), 
+                page.ClickAsync("text=Users"));
+
+            Assert.IsTrue(response.Ok); // This fails because the port for the API is hard coded. It doesn't ever hit the API in WebHostServerFixture.
```
breaman, line 62: The api port is not hard coded (or at least it shouldn't be based on the fact we made it a variable in the appsettings.json file. This then gets modified in our WebHostServerFixture<WStartup, AStartup> class when we launch things. Navigation should work just fine for the page as well, regardless of if the API can load data or not.

SecretSanta/test/SecretSanta.E2ETests/EndToEndTests.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                Headless = true
+            });
+
+            var page = await browser.NewPageAsync();
+            var response = await page.GoToAsync(localhost);
+
+            Assert.IsTrue(response.Ok);
+            
+            await Task.WhenAll(
```
breaman, line 57: You should not need this Task.WhenAll call. To navigate, you just need to do something like the following
```
await page.ClickAsync("text=Users");
var button = await page.WaitForSelectorAsync("a:has-text('Create User')");
```
I do like what you were attempting to do here, though.

breaman (commented):
Overall, nice work on the assignment. Didn't really have code to correct, just made some comments around your comments.

PR #176, opened 2021-05-17
Assignment 7

CalumMackenzie-EWU:


Thunderbird145:
I will review this.

Thunderbird145:

Everything that's done so far looks okay. Just need to finish it up.

breaman (commented):
Nice work on the assignment. I do enjoy reading your comments as well. I know it doesn't add to the readability of the code (which means you might not want to do this outside of the academic world) but it is nice to see your learning process.

PR #177, opened 2021-05-17
Assignment7 completed

Thunderbird145:
Managed to get the E2E done, but haven't even started on the alpine JS


NicholasJohnsonDev:
Your title says unfinished but your last commit says completed, so I'm gonna guess it's ready for review.
I'll review this...

NicholasJohnsonDev:
Everything builds and runs. All Playwright E2E tests pass and each one has a picture too. CRUD still works, handled with AlpineJS.
Looks great!

Thunderbird145:
I have changed the title. Thanks for the review!

SecretSanta/src/SecretSanta.Web/scripts/main.ts
```diff
@@ -6,10 +6,15 @@ import { library, dom } from "@fortawesome/fontawesome-svg-core";
 import { fas } from '@fortawesome/free-solid-svg-icons';
 import { far } from '@fortawesome/free-regular-svg-icons';
 import { fab } from '@fortawesome/free-brands-svg-icons';
+import axios from 'axios';
+
+import {User, UsersClient} from '../Api/SecretSanta.Api.Client.g';
```
breaman, line 11: I see you used the generated client for typescript (which was not covered in class), nice touch. Was it a little clearer to go this route instead of what was covered in class?

SecretSanta/test/SecretSanta.E2E.Tests/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,205 @@
...
+            });
+
+            var page = await browser.NewPageAsync();
+            var response = await page.GoToAsync(localhost);
+
+            Assert.IsTrue(response.Ok);
+
+            await page.ClickAsync("text=Users");
+
+            await page.ScreenshotAsync("users.png");
```
breaman, line 55: This isn't a great way to verify that navigation occurred since this would still require some human intervention to look at the screenshots. Most of the time you would want to have some type of Assert to make sure the navigation occurred.

breaman (commented):
Nice job on the assignment, hit all the points that were asked for.

PR #180, opened 2021-05-17
Assignment7work

thecactus1:


MeguminXIII:
I will review

MeguminXIII (commented):
Assignment

It looks like you have yet to do the alpine.js stuff yet.  Other than that it looks really good. nice job :)

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                SlowMo = 250
+            });
+
+            var page = await browser.NewPageAsync();
+            var response = await page.GoToAsync(localhost);
+
+            Assert.IsTrue(response.Ok);
+
+            var Header = await page.GetTextContentAsync("body > header > div > a");
+            Assert.AreEqual("SecretSanta", Header);
```
breaman, line 34: This test fails since the title is Secret Santa, not SecretSanta.

SecretSanta/src/SecretSanta.Web/Startup.cs
```diff
@@ -20,6 +20,14 @@ public void ConfigureServices(IServiceCollection services)
         {
```
breaman, line 20: BaseAddress should not have been hardcoded here, it was suppose to be pulled from the configuration/appSettings file.

SecretSanta/test/SecretSanta.E2E/EndToEnd/EndToEndTests.cs
```diff
@@ -0,0 +1,199 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Threading.Tasks;
+using PlaywrightSharp;
+using System.Linq;
+
+namespace SecretSanta.Web.Tests{
+    [TestClass]
```
breaman, line 7: None of these tests passed, my guess is it has something to do with the fact that the api host is hard coded, but other elements that should not have relied on the api also timed out.

breaman (commented):
The AlpineJS stuff looked good. I noticed you used the generated typescript client, probably should have generated it in the api project like we did with the cs stuff so if the API changes, typescript would automatically get the updates as well.

PR #182, opened 2021-05-23
Completed Assignment 8.

delta-plus:
Done!

### DIFF
--- a/SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupsControllerTests.cs
+++ b/SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupsControllerTests.cs
@@ -222,7 +222,7 @@ namespace SecretSanta.Api.Tests.Controllers
         }
 
         [TestMethod]
-        public async Task CreateAssignments_WithSuccess_ReturnsOk()
+        public async Task Assign_WithSuccess_ReturnsOk()
         {
             //Arrange
             using WebApplicationFactory factory = new();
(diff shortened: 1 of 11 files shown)


jamesgreiner:
I will review this.

jamesgreiner (commented):
Write appropriate unit tests for changes made to the GroupRepository ✔
Write appropriate functional tests for the change made to the GroupsController ✔
Extra Credit


Looks good! All the required specs are complete.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupsControllerTests.cs
```diff
@@ -262,13 +263,10 @@ public async Task CreateAssignments_WithError_ReturnsBadRequest()
             HttpClient client = factory.CreateClient();
 
             //Act
-            HttpResponseMessage response = await client.PutAsJsonAsync("/api/groups/42/assign", new Dto.UpdateGroup
-            {
-                Name = "Changed"
-            });
+            HttpResponseMessage response = await client.PostAsJsonAsync("/api/groups/42/assign", 42);
```
Keboo, line 266: nit: Your post method does not read anything from the body, so you don't need to pass a value. You can just pass nothing.
```suggestion
            HttpResponseMessage response = await client.PostAsJsonAsync("/api/groups/42/assign", "");
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupsControllerTests.cs
```diff
@@ -222,7 +222,7 @@ public async Task Update_InvalidGroupId_ReturnsNotFound()
         }
 
         [TestMethod]
-        public async Task CreateAssignments_WithSuccess_ReturnsOk()
+        public async Task Assign_WithSuccess_ReturnsOk()
```
Keboo, line 225: These test work, but it only checks the return value. It would be good to also assert that this endpoint invokes the right GroupRepository method.

Keboo (commented):
# Assignment

PR #184, opened 2021-05-24
Assignment8

ZStuefen:


Thunderbird145:
I will review.

Thunderbird145:
# Assignment

## Extra Credit

Thunderbird145:
looks like the unit tests are missing. Everything else appears to be present but I was unable to actually test it, running the website results in everything from the api server coming in as undefined. If it works on your end this might just be an issue I'm having, because everything looks okay as far as I can tell.

SecretSanta/src/SecretSanta.Api/Controllers/GroupsController.cs
```diff
@@ -109,5 +109,19 @@ public ActionResult Add(int id, [FromBody] int userId)
             }
             return NotFound();
         }
+
+        [HttpPost("{id}/assign")]
+        [ProducesResponseType((int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
```
Keboo, line 115: The method does not appear to return a NotFound.
```suggestion
```

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -45,5 +45,25 @@ public void Save(Group item)
...
+        public AssignmentResult GenerateAssignments(int id)
+        {
+            bool res = MockData.Groups.TryGetValue(id, out Group? group);
+            if (res == false || group is null || group.Name is null || group.Name.Length == 0)
+            {
+                return AssignmentResult.Error("Group not found");
+            }
+            if (group.Users.Count <= 2)
+            {
+                return AssignmentResult.Error("Group " + group.Name + " must have at least three users");
```
Keboo, line 58: C# has a cleaner way to build up strings called string interpolation.
```suggestion
                return AssignmentResult.Error($"Group {group.Name} must have at least three users");
```

Keboo (commented):
# Assignment

PR #187, opened 2021-05-24
Assignment8

NicholasJohnsonDev:
Assignment 8


NicholasJohnsonDev:
Currently working on this, should be done by 2pm

Fervidusletum:
Reviewing

Fervidusletum (changes requested):
## Assignment

## Extra Credit

## Comments
- The GroupRepository GenerateAssignments_WithInvalidId test intermittently fails.
  - I ran into a similar issue; I fixed it by changing the test value from 42 to a unique value.
- Both GroupController tests for CreateAssignments are failing.
  - WithError expects a badrequest, but gets a notfound.
  - WithSuccess expects an OK but gets a notfound.
- GenerateAssignments isn't tested with regard to ordering of generated assignments.
- While the alert works for displaying errors, it's rather annoying. IMO it would be better to display the error in-line on the page.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -36,6 +37,51 @@ public bool Remove(int id)
...
+        public AssignmentResult GenerateAssignments(int id)
+        {
+
+            if (!MockData.Groups.ContainsKey(id))
+            {
+                return AssignmentResult.Error("Group not found");
+            }
+            else if (MockData.Groups[id].Users.Count() <= 2)
+            {
+                return AssignmentResult.Error("Group Group must have at least three users");
```
Keboo, line 64:
```suggestion
                return AssignmentResult.Error("Group must have at least three users");
```

SecretSanta/test/SecretSanta.Business.Tests/GroupRepositoryTests.cs
```diff
@@ -122,6 +122,26 @@ public void GenerateAssignments_WithInvalidId_ReturnsError()
             Assert.AreEqual("Group not found", result.ErrorMessage);
         }
 
+        [TestMethod]
+        public void GenerateAssignments_WithInvalidId_CreatesAssignments()
```
Keboo, line 126:
```suggestion
        public void GenerateAssignments_WithInvalidId_ReturnsNotFoundError()
```

Keboo (commented):
# Assignment

Keboo:
# Assignment

PR #188, opened 2021-05-24
Assignment8 @anthonehi

roylalog000:
anthonehi


jmplett:
I will review

OkayNiceBrian (commented):
Write appropriate functional tests for the change made to the GroupsController ❌

No added tests to test the functionality.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -35,7 +42,45 @@ public bool Remove(int id)
...
 
+            for(int i = 0;i < group.Users.Count;i++)
+            {
+                User randUser = userList[random.Next(userList.Count)];
+                while(randUser.Id == group.Users[i].Id)
+                {
+                    randUser = userList[random.Next(userList.Count)];
+                }
+                userList.Remove(randUser);
+                group.Assignments.Add(new Assignment(group.Users[i],randUser));
```
Keboo, line 76: When I run it the `group.Users[i]` here ends up throwing an IndexOutOfRangeException. I believe this is because of the issue mentioned above where users are being removed from the group.

Keboo (commented):
# Assignment
  - A means to display any errors that occurs while generating assignments. ❌- Though error messages are being passed back to the client, they appear to be getting ignored.

It appears the test projects are failing to compile due to the interface changes. It also appears that generating assignments always fails with an IndexOutOfRangeException.

PR #190, opened 2021-05-24
Assignment 8

MeguminXIII:


ZStuefen:
I will review this

ZStuefen:
The unit tests did not work automatically because you named the new method Assign() rather than GenerateAssignments()
I changed the tests to test that method.

# Assignment
      - The error is created but Groups/Edit does not have the functionality to display it
  - A means to generate assignments. ❌
  - Display of the assignments for each of the users. ❌
      Groups/Edit was not changed
    I'll check these off because the tests will still cover the Assign method if changed to use Assign() instead of GenerateAssignments(). One test does not pass due to a typo in "Groups must ahve at least 3 users"

## Extra Credit

SecretSanta/src/SecretSanta.Web/Views/Groups/Edit.cshtml
```diff
@@ -40,5 +43,8 @@
         </select>
         <button class="border-1 bg-green-200 border-green-600 hover:bg-green-400 py-2 px-4 rounded-xl mr-3" @@click="addToGroup(group.id)">Add</button>
     </div>
+    <div class="py-4">
+        <button class="border-1 bg-green-400 border-black-800 hover:bg-red-600 py-2 px-4 rounded-xl mr-3" @@click="Assign(group.id)">Generate Assignments</button>
```
Keboo, line 47: The `Assign` method referenced here does not exist in the main.ts so clicking the button results in an error.

Keboo (commented):
# Assignment
  - A means to generate assignments. ❌
  - A means to display any errors that occurs while generating assignments. ✔❌ - this is sort of there, but given the issue generating assignments it is difficult to test.

PR #191, opened 2021-05-24
Assignment 8

TechGlue:
Working in progress still missing most of the features.

### DIFF
--- a/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
+++ b/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
@@ -50,6 +50,7 @@ namespace SecretSanta.Business
 
         public AssignmentResult GenerateAssignments(int id)
         {
+            //grab and do some checks
             Group group = GetItem(id);
 
             if (group is null)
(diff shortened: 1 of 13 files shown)


Ashrom0:
I will review this

Ashrom0 (commented):
- Update the GroupRepository with functionality to generate gift assignments. ✔
- Write appropriate unit tests for changes made to the GroupRepository ❌
- Write appropriate functional tests for the change made to the GroupsController ❌

Seems that the code fits in with what the Assignment is asking, but for tests, not able to get them to work, not sure if its my part or the users unchanged code, will modify update if that is the first case.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -45,5 +47,53 @@ public void Save(Group item)
...
         }
+
+        public AssignmentResult GenerateAssignments(int id)
+        {
+            //grab and do some checks
+            Group group = GetItem(id);
+
+            if (group is null)
+            {
+                throw new NullReferenceException();
```
Keboo, line 58: nit: This is the wrong exception type to throw here. `NullReferenceException` is not one that you are ever expected to throw yourself. Better options would be `InvalidOperationException` or perhaps `ObjectNotFoundException`

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -45,5 +47,53 @@ public void Save(Group item)
...
+            return AssignmentResult.Success();
+        }
+
+        //list shuffling method from stack overflow
+        //source - https://stackoverflow.com/questions/273313/randomize-a-listt
+        private void Shuffle<T>(IList<T> list)
+        {
+            if (list == null)
+            {
+                throw new NullReferenceException();
```
Keboo, line 85: In this case the correct exception to throw would be one of the argument exceptions since this is passed as an argument.

```suggestion
                throw new ArgumentNullException(nameof(list));
```

Keboo (commented):
# Assignment

PR #192, opened 2021-05-24
Assignment8

Cybotive:


CalumMackenzie-EWU:
Starting review

SecretSanta/src/SecretSanta.Api/Controllers/GroupsController.cs
```diff
@@ -109,5 +109,20 @@ public ActionResult Add(int id, [FromBody] int userId)
...
+        [HttpPut("{id}/generateAssignments")]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        [ProducesResponseType((int)HttpStatusCode.NoContent)]
+        public ActionResult GenerateAssignments(int id)
+        {
+            var result = GroupRepository.GenerateAssignments(id);
+            
+            if (result.IsSuccess)
+            {
+                return NoContent();
```
CalumMackenzie-EWU, line 122: I dont know if this is the best thing to return for the success case. But if it works, then all good.

CalumMackenzie-EWU (commented):
Sorry I couldnt get your Api project to build, but that could be on my end. For some reason Ive been having a lot of issues when trying to open projects with vs code. So Im basing everything off of what I can see which isnt the best representation, again sorry.

Display of the assignments for each of the users. ❌
Write appropriate unit tests for changes made to the GroupRepository ✔
Write appropriate functional tests for the change made to the GroupsController ✔
Extra Credit

For the X's I couldn't see any changes to your Edit page in the files changed section of github. So I assume they havent been pushed yet.

SecretSanta/src/SecretSanta.Api/Controllers/GroupsController.cs
Cybotive: It is because I believe there is no body in the HTTP response. However I could be wrong about that.

Keboo: Agreed, NoContent (204) is perfectly fine when there is no content.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -45,5 +55,70 @@ public void Save(Group item)
...
+                    Assignment assignment = new(users[i], users[recipientIndexes[i]]);
+                    tempAssignments.Add(assignment);
+                }
+
+                // GRADER: A user is not allowed to be both the Giver and Recipient of the assignment.
+                foreach (Assignment assign in tempAssignments)
+                {
+                    if (assign.Giver == assign.Receiver) // Equal by reference is intentional.
+                    {
+                        return AssignmentResult.Error("Unable to generate assignments");
```
Keboo, line 99: This works, but it would be nice if the algorithm handled the error case and either retried or addressed so this case does not occur.

SecretSanta/test/SecretSanta.Business.Tests/GroupRepositoryTests.cs
```diff
@@ -158,5 +160,57 @@ public void GenerateAssignments_WithValidGroup_CreatesAssignments()
             Assert.AreEqual(3, group.Assignments.Select(x => x.Receiver.FirstName).Distinct().Count());
             Assert.IsFalse(group.Assignments.Any(x => x.Giver.FirstName == x.Receiver.FirstName));
         }
+
+        [TestMethod]
+        public void GenerateAssignments_GivenRandom_UsesInstance()
```
Keboo, line 165: This works, and I like how you handled the `Random` for the purpose of testing. Though it appears the error case for `"Unable to generate assignments"` does not get hit in any of these tests.

Keboo (commented):
# Assignment

PR #193, opened 2021-05-24
Assignment8

jamesgreiner:
Just made an initial commit, will work on this over the next two days.

### DIFF
--- a/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
+++ b/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
@@ -19,6 +19,11 @@ namespace SecretSanta.Business
             return item;
         }
 
+        public AssignmentResult GenerateGiftAssignment(int id)
+        {
+            
+        }
+
         public Group? GetItem(int id)
         {
             if (MockData.Groups.TryGetValue(id, out Group? user))
--- a/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
+++ b/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using SecretSanta.Data;
 
+//A group with with 2 or fewer users should result in an error. This error should be displayed to a user.
+
 namespace SecretSanta.Business
 {
     public class GroupRepository : IGroupRepository
--- a/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
+++ b/SecretSanta/src/SecretSanta.Business/GroupRepository.cs
@@ -2,12 +2,11 @@
 using System.Collections.Generic;
 using SecretSanta.Data;
 
-//A group with with 2 or fewer users should result in an error. This error should be displayed to a user.
-
 namespace SecretSanta.Business
 {
     public class GroupRepository : IGroupRepository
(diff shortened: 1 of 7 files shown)


NicholasJohnsonDev:
It doesn't look like you've finished. If you finish by tomorrow evening, tag me and I'll give it another review.

Assignment
Update the GroupRepository with functionality to generate gift assignments. ❌
A group with with 2 or fewer users should result in an error. This error should be displayed to a user. ❌
need to implement this. just gotta check if the user is already a giver/recipient and throw an error if it is.

Update the GroupsController with an endpoint for generating the assignments. ❌
Display of the assignments for each of the users. ❌

Write appropriate functional tests for the change made to the GroupsControllers❌
need to implement

Extra Credit
Allow for navigating to a user from the Group edit page to make it easy to see the gifts requested by a user.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -45,5 +73,19 @@ public void Save(Group item)
 
             MockData.Groups[item.Id] = item;
         }
+
+       //Link to the location of a shuffle method that I found: https://stackoverflow.com/questions/273313/randomize-a-listt
+       private void Shuffle<T>(this IList<T> list)
```
Keboo, line 78: This is declared like an [extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods), however extension methods must live inside of static classes (hence the current compilation error). However, since you are just using it like a normal method you can simply drop the `this` in front of the parameter.
```suggestion
       private void Shuffle<T>(IList<T> list)  
```

SecretSanta/src/SecretSanta.Web/Views/Groups/Edit.cshtml
```diff
@@ -1,4 +1,7 @@
-<section x-data="SecretSanta.createOrUpdateGroup()" x-init="loadData()">
+<section x-data="SecretSanta.changeGroup()" x-init="loadData()">
```
Keboo, line 1: This is referencing `SecretSanta.changeGroup()` but in the main.ts the function is still called `createOrUpdateGroup()` this is causing errors when loading the page.

SecretSanta/src/SecretSanta.Web/scripts/main.ts
```diff
@@ -189,6 +241,30 @@ export function createOrUpdateGroup() {
                 console.log(error);
             }
             await this.loadGroup();
+        },
+
+        async assignUsersWithinGroup(currentGroupId: number) {
```
Keboo, line 246: This method is not implemented so it is difficult to test the assignment code.

Keboo (commented):
# Assignment
- Update the `GroupRepository` with functionality to generate gift assignments. ❌ - Current project doesn't compile due to the Shuffle method being declared like an extension method.
  - A means to generate assignments. ❌ - assignUsersWithinGroup lacks an implementation.
- Write appropriate unit tests for changes made to the `GroupRepository` ❌ - Current assignment tests are failing.
- Write appropriate functional tests for the change made to the `GroupsController` ❌ - Current assignment tests are failing.

PR #194, opened 2021-05-24
Assignment 8.

CalumMackenzie-EWU:
I think everything's working now except for the two tests for an ok response and bad request.


jmplett:
I will review

jmplett:
Update the GroupsController with an endpoint for generating the assignments. ❌
Display of the assignments for each of the users. ❌ (sort of done i cant see the actual assignments)
Write appropriate functional tests for the change made to the GroupsController ❌
Extra Credit


Good start to the assignment as you have started building up a good framework for the completion of the assignment. Just need to finish the implementation of assigning and such. If you would like me to review this again at a later time just let me know and I would gladly do that for you.

Keboo (commented):
# Assignment

PR #196, opened 2021-05-24
Assignment 8

Fervidusletum:


MeguminXIII:
I will review

MeguminXIII (commented):
Assignment
need to implement this. just gotta check if the user is already a giver/recipient and throw an error if it is.

Write appropriate functional tests for the change made to the GroupsControllers❌
need to implement

Extra Credit

looks good, need to implement tests and put in the functionality for making it so a user cannot be both a giver and a recipient. Otherwise it looks good.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -45,5 +52,30 @@ public void Save(Group item)
...
+        {
+            Group? group = GetItem(id);
+            if (group?.Users is null) return AssignmentResult.Error("Group not found");
+            if (group.Users.Count < 3) return AssignmentResult.Error($"Group {group?.Name} must have at least three users");
+
+            group.Assignments.Clear();
+
+            
+            List<User> shuffled = ShuffleList(group.Users);
+            shuffled.Add(shuffled.First()); // add first user to end too, so we dont need extra circular link logic
```
Keboo, line 66: Clever implementation

Keboo (commented):
# Assignment

PR #197, opened 2021-05-25
Assignment8

OkayNiceBrian:


thecactus1:
Reviewing

thecactus1:
In this assignment we are going to build out some of the logic needed for Secret Santa.
There have been several updates to the SecretSanta repository.
Key changes to be aware of:
- The data project has been expanded and contains users for groups and a new Assignment class.
- There is a new `GroupRepository` added to the business project.
- The gifts section links has been removed from the web project. The pages are still present but should not be needed.
- Opening the "edit" page of a group now also shows the users in a group, and has separate controls for editing the group and adding and removing users from the group.
- No additional Web or E2E tests are needed for this assignment.

# Assignment
  - A group with with 2 or fewer users should result in an error. This error should be displayed to a user. ❌

## Extra Credit

None of the tests functioned. In API, I'm not sure why, but dotnet test caused a hang. In Business, Dotnet test failed to compile because not all the interface members were implemented in the testable group repository.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -36,6 +36,44 @@ public bool Remove(int id)
...
+                group.Assignments.Clear();
+            }
+
+            List<User> users = new List<User>(group.Users);
+            System.Random random = new System.Random();
+
+            for (int i = 0; i < group.Users.Count; i++) 
+            {
+                User rUser = users[random.Next(users.Count)];
+                while (rUser.Id == group.Users[i].Id)
```
Keboo, line 66: There is a possible infinite loop here. Imagine a group with 3 users A, B, and C. Lets assume
A randomly gets assigned B, and B randomly gets assigned A. This leaves only user C left in your `users` list and it will loop forever.

SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -46,4 +84,26 @@ public void Save(Group item)
...
+        public string? ErrorMessage { get; }
+
+        private AddAssignmentResult(string? errorMessage)
+        {
+            ErrorMessage = errorMessage;
+        }
+
+        public static AddAssignmentResult UserNotFound()
+            => new AddAssignmentResult("User not found");
+        public static AddAssignmentResult Error(string? eMessage)
```
Keboo, line 100: nit: In general we avoid abbreviations unless they are well know.
```suggestion
        public static AddAssignmentResult Error(string? errorMessage)
```

Keboo (commented):
# Assignment
  - A means to display any errors that occurs while generating assignments. ❌ - The errors are returned but they are not displayed anywhere.
- Write appropriate functional tests for the change made to the `GroupsController` ❌ - Test project currently does not compile since TestableGroupRepository was not updated.

PR #199, opened 2021-05-30
Assignment9-johnson-m

thecactus1:
Right now, I get this error when migrating or running tests. It's totally alien, as Usergroup has no equivalent.
image


SecretSanta/src/SecretSanta.Api/Program.cs
```diff
@@ -1,15 +1,45 @@
...
+            string deploy = "";
+            if(args.Length > 0){
+                deploy = args[0];
+            }
+            if(deploy == "--DeploySampleData"){
+
+                Console.WriteLine("Deploying Data...");
+                using (var dbcontext = new DbContext()){
+                    dbcontext.Database.EnsureDeleted();
+                    dbcontext.Database.EnsureCreated();
```
Fervidusletum, line 22: You should use `dbcontext.Database.Migrate()` instead of `.EnsureCreated()`
EnsureCreated prevents migrations from being applied.

Fervidusletum (commented):
## Assignment

1. Configure all data to be stored in an SqLite Database using the Entity Framework. ✔
    - All tables must have a primary key ✔
    - All tables mush have an unique/alternate key (other than the primary key) ❌

2. Provide UI functionality for a complete Secret Santa functionality. ❌

3. Ensure website is deployed by your GitHub Action and proivde URL to the deployed site in your pull request. ❌

4. Provide a walk through video (or do the extra credit) showing: ❌

5. Remove the MockData class and replace it with sample data that gets deployed into the database following a migration when the command "DeploySampleData" is specified on the command line. ✔

6. Add some minimal rudimentary logging using `Microsoft.Extensions.Logging`. 
    - At a minimum, the logging should be in startup.cs. ❌
    - Configure the Entity Framework to log to a file (located in the SecretSanta.Data assembly directory) with the category name "Database". ✔
    - The name of the log file should default to `db.log`. ✔
    - The name of the log file should be configurable using `Microsoft.Extensions.Coniguration`. ❌

7. Allow for the connection string (db path) for the SqLite database to be provided using `Microsoft.Extensions.Coniguration`:
    - Default to main.db in the same directory as the SecretSanta.Data assembly. ✔
    - Allow for an environment variable to override the default connection string ❌
    - Allow a connection string passed on the command line to take precidence over the environment variable. ❌

9. GitHub Action CI/CD Build ❌

## Comments
 - Unable to check for UI functionality due to DB issues, though the gift CRUD code looked fine at a glance
 - Assignment does not have an alternate key
  - you can add the properties `public int GiverId {get; set;}` and `public int ReceiverId {get; set;}` to use for an alternate key.  EF will autopopulate their values from `Giver` and `Receiver`.
 - you might encounter foreign key constraint issues when updating the value of an alternate key, such as updating a user's name.  I got around that by specifying unique attributes instead of alternate keys.
 - you can add serilog in startup with some minimal setup and it will auto propagate through most of the backend

PR #200, opened 2021-05-31
Assignment 9 Best I could do (F5 implemented with studio code.)

Thunderbird145:


SecretSanta/src/SecretSanta.Business/GroupRepository.cs
```diff
@@ -13,37 +14,75 @@ public Group Create(Group item)
                 throw new ArgumentNullException(nameof(item));
             }
 
-            MockData.Groups[item.Id] = item;
+            using var DatabaseContext = new DatabaseContext();
```
mmwoodfo, line 17:
```suggestion
            using var DatabaseContext = new new DbContext();
```
Is this what you meant to do here? There are 60 errors in your code because it thinks "DatabaseContext" is the name of a variable.

Thunderbird145: It should be using var DatabaseContext. My last push did not go through last night, I fixed anything like this in that i think

mmwoodfo: It looks like PR #215 still has this issue, although the issue is now :
`using var Context = new Context();` 
instead of `using var Context = new DbContext();`

PR #202, opened 2021-06-04
Assignment9

Cybotive:
[**Build and Test Workflow**](https://github.com/Cybotive/EWU-CSCD379-2021-Spring/actions/workflows/AssignmentBuildTest.yml)

[**Link to deployed website**](https://cybotive-ewu-cscd379-2021-spring.azurewebsites.net)

[**Link to my branch on my fork**](https://github.com/Cybotive/EWU-CSCD379-2021-Spring/tree/Assignment9)

Demonstration:
https://youtu.be/52Tt2ZgCyy8

Forgot to show Gift Create page that you can get to via editing a user. But adding a gift fails on the API side anyway (but the front end is completely functional with a receiver user picker and everything).

I had alternate keys but turned them off during development. Then when I turned them back on all my tests failed. But the code's there.


jmplett:
I will review

jmplett:
# Assignment 9 & 10

## Overview

In this assignment we are going to build "complete" the SecretSanta project. As discussed in class on Thursday, you have two weeks to complete the assignment which is worth "double the points".

## Assignment

- Configure all data to be stored in an SqLite Database using the Entity Framework. ✔
   - Hint: Add a Gift object to the SecretSanta.Data project
   - Hint: Update `Group` and `Assignment` to persist as a many-to-many relationship
   - All tables must have a primary key
  - All tables mush have an unique/alternate key (other than the primary key). _**(I don't think you met this specification based on what you have I don't think something like a title/name alone would work as you could get many of the same)**_
- Provide UI functionality for a complete Secret Santa functionality. ❌
 This includes:
  - The ability for a user to add a list of gifts
   - Displaying who a user's secret santa is for in each group.✔
   - Viewing all gifts requested by your secret santa
  - Provide an end-to-end test that:
   - Creates a Group
   - Creates a 3 users
   - Assigns a minimum of two gifts to each user
   - Assigns the secret santas
   - Allows a user to see their secret santa and the associated gifts wanted.
 - Remove the MockData class and replace it with sample data that gets deployed into the database following a migration when the command "DeploySampleData" is specified on the command line. _**(Started just missing seed data)**_
 - Allow for the connection string (db path) for the SqLite database to be provided using `Microsoft.Extensions.Coniguration`: ❌
   - Default to main.db in the same directory as the SecretSanta.Data assembly.
   - Allow for an environment variable called to override the defaiult connection string
   - Allow a connection string passed on the command line to take precidence over the environment variable.
 - Add some minimal rudimentary logging using `Microsoft.Extensions.Logging`. At a minimum, the logging should be in startup.cs ❌
 - Configure the Entity Framework to log to a file (located in the SecretSanta.Data assembly directory) with the category name "Database". The name of the log file should be conifgurable but default to db.log.
- GitHub Action CI/CD Build 
   - Ensure all tests (unit test, end to end tests, etc.) are running in your GitHub Action build and include a URL to the build in your executing GitHub action in your pull request. _**Even though not complete you have completed this step at the current state**_ ✔
   - Ensure website is deployed by your GitHub Action and proivde URL to the deployed site in your pull request. _**(not deployed but like available)**_ ❌

Cybotive:
"All tables mush have an unique/alternate key (other than the primary key). (I don't think you met this specification based on what you have I don't think something like a title/name alone would work as you could get many of the same) ❌"
Yeah, this has been revealing a bug in my context setup as well. Just not sure how I could create a unique alternate key that a user would know how to use.

PR #204, opened 2021-06-06
Assignment9

delta-plus:
Well, you can add a list of gifts to users now, which is nice.


MeguminXIII:
will review

SecretSanta/src/SecretSanta.Data/DbContext.cs
```diff
@@ -0,0 +1,200 @@
...
+              {
+                  Id = 10,
+                  GroupId = 2,
+                  UserId = 5
+              },
+            };
+
+            modelBuilder.Entity<User>().HasData(users);
+            modelBuilder.Entity<Group>().HasData(groups);
+            modelBuilder.Entity<GroupUser>().HasData(groupUsers); */
```
MeguminXIII, line 197: Probably don't need this since you have a data seeder class

SecretSanta/src/SecretSanta.Data/DbContext.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            Database.Migrate();
+        }
+
+        public Microsoft.EntityFrameworkCore.DbSet<User> Users => Set<User>();
+        public Microsoft.EntityFrameworkCore.DbSet<Group> Groups => Set<Group>();
+        public Microsoft.EntityFrameworkCore.DbSet<Gift> Gifts => Set<Gift>();
+        public Microsoft.EntityFrameworkCore.DbSet<Assignment> Assignments => Set<Assignment>();
+//        public Microsoft.EntityFrameworkCore.DbSet<GroupUser> GroupUsers => Set<GroupUser>();
+//        public Microsoft.EntityFrameworkCore.DbSet<GroupAssignment> GroupAssignments => Set<GroupAssignment>();
```
MeguminXIII, line 26: can probably delete these, unless you plan on using them later?

MeguminXIII (commented):
Assignment✔❌
Configure all data to be stored in an SqLite Database using the Entity Framework. ✔
Hint: Add a Gift object to the SecretSanta.Data project ✔
Hint: Update Group and Assignment to persist as a many-to-many relationship✔
All tables must have a primary key✔
All tables mush have an unique/alternate key (other than the primary key)✔

Provide UI functionality for a complete Secret Santa functionality. This includes:❌
The ability for a user to add a list of gifts❌ not there yet
Displaying who a user's secret santa is for in each group. ❌ not there yet
Viewing all gifts requested by your secret santa ❌ not there yet

Provide an end-to-end test that: ❌ no tests yet that I see
Creates a Group❌
Creates a 3 users❌
Assigns a minimum of two gifts to each user❌
Assigns the secret santas❌
Allows a user to see their secret santa and the associated gifts wanted.❌

Remove the MockData class and replace it with sample data that gets deployed into the database following a migration when the command "DeploySampleData" is specified on the command line.❌
I see that you have replaced MockData with a DataSeeder.Seed() method. So I think you need to take in the args[0] and check to see if DeploySampleData is specified as an argument when you run the program? or maybe there's another way to read in from the terminal while the program is running that I can't think of right now. The only way I can think of would be a prompt and scanning in the command, but that would halt the program until a command is entered so I don't think that would work. So I think it has to mean something like dotnet run ProgramName DeploySampleData, I think that's what it means at least.

Allow for the connection string (db path) for the SqLite database to be provided using Microsoft.Extensions.Coniguration:❌
Default to main.db in the same directory as the SecretSanta.Data assembly.❌
Allow for an environment variable called to override the defaiult connection string❌
Allow a connection string passed on the command line to take precidence over the environment variable.❌
Not implemented yet, I guess a suggestion would be then to take in args[1] as well? and check to make sure it exists before trying to access it of course.

Add some minimal rudimentary logging using Microsoft.Extensions.Logging. At a minimum, the logging should be in startup.cs❌
Looks like you have some logging in DbContext, but its not done with the Microsoft.Extensions.Logging and it does specify startup.cs which it looks like you haven't done. 

Configure the Entity Framework to log to a file (located in the SecretSanta.Data assembly directory) with the category name "Database". The name of the log file should be conifgurable but default to db.log.❌
Not implemented yet

GitHub Action CI/CD Build
Ensure all tests (unit test, end to end tests, etc.) are running in your GitHub Action build and include a URL to the build in your executing GitHub action in your pull request.❌
Ensure website is deployed by your GitHub Action and proivde URL to the deployed site in your pull request.❌

Looks pretty good so far, just have to implement stuff. Good job :)

PR #209, opened 2021-06-07
Assignment 9

CalumMackenzie-EWU:
Sadly didnt get things working. The website can only add Groups to the db, but not Users. It has something to do with my Post since I can see it returning a 400 inside the devtools. But didnt figure out how to resolve in time. Also unsure if this is just on my end, but the Api project can be run, but it gets hung up on the command line, still functions to be able to pull data from the db and create the log file.


CalumMackenzie-EWU:
Still working to get database set up. No web interaction yet so no need to run the web project. Running the Api will just seed the database inside the .Data folder. Please let me know if that actually work on your end cause I'm using some directory setting stuff which can cause possible issues. Basically run the api project, and then close it. Then view the database to see if it populated on your end, thank you.

PR #212, opened 2021-06-07
Assignment 9

TechGlue:
Work in progress. No review needed. Pair programming with Pedro Hernandez (@Ashrom0).


TechGlue:
Stopped on linking Group repository with DbContext.

TechGlue:
Final commit to make grading easier. I spent too much time trying to fix issues with the table and the collision of keys. Too much time spent trying to get dotnet ef to run. Most of my tests are failing, barely even touched the feature that we had to implement that being gifts. Did not get the F5 setup command running because I program on Rider and with all the issues going on my confidence just dropped. I'll continue and learn the tools to get the job done but for now, this is all I was able to accomplish. Thank you, for a great quarter guys.

